PR #1, opened 2026-04-30
Add test suite (75 tests) + fix platform-neutral install hints

gatosa12:
## What this PR does

This is the first test suite for the project, plus two bug fixes for Linux/Windows users.

### New: `tests/` — 75 unit tests (no external binaries required)

| File | Tests | Covers |
|---|---|---|
| `tests/test_transcribe.py` | 22 | `parse_vtt`, `_dedupe`, `filter_range`, `format_transcript` |
| `tests/test_frames.py` | 25 | `parse_time`, `format_time`, `_clamp_fps`, `auto_fps`, `auto_fps_focus` |
| `tests/test_download.py` | 28 | `is_url`, `resolve_local`, `_pick_subtitle`, `_pick_video`, `VIDEO_EXTS` |

All tests run with just `pip install pytest` — no ffmpeg, yt-dlp, or API keys needed.

### New: `.github/workflows/tests.yml` — CI workflow

Runs the test suite on every push/PR against Python 3.10, 3.11, and 3.12.

### Bug fixes: platform-neutral install hints

**`scripts/download.py`** — the error message when `yt-dlp` isn't found previously said `brew install yt-dlp` on all platforms. Fixed to show the right command per OS.

**`scripts/frames.py`** — same issue in two places for `ffmpeg`/`ffprobe`. Fixed to show macOS / Linux / Windows install commands.

### How to run tests locally

```bash
pip install pytest
pytest tests/ -v
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,30 @@
+name: Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  test:
+    name: pytest (Python ${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: pip install pytest
+
+      - name: Run tests
+        run: pytest tests/ -v --tb=short
--- a/scripts/download.py
+++ b/scripts/download.py
@@ -59,7 +59,10 @@ def _pick_video(out_dir: Path) -> Path | None:
 
 def download_url(url: str, out_dir: Path) -> dict:
     if shutil.which("yt-dlp") is None:
-        raise SystemExit("yt-dlp is not installed. Install with: brew install yt-dlp")
+        raise SystemExit(
+            "yt-dlp is not installed. "
+            "macOS: brew install yt-dlp  |  Linux/Windows: pip install yt-dlp"
+        )
 
     out_dir.mkdir(parents=True, exist_ok=True)
     output_template = str(out_dir / "video.%(ext)s")
--- a/scripts/frames.py
+++ b/scripts/frames.py
@@ -19,12 +19,10 @@ MAX_FPS = 2.0
 
 
 def _clamp_fps(fps: float, duration_seconds: float, max_frames: int) -> tuple[float, int]:
-        intended = int(round(fps * duration_seconds))
-    fps = min(fps, MAX_FPS)
-    target = min(max_frames, max(1, max(intended, int(round(fps * duration_seconds)))))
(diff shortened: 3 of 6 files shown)


gatosa12 (commented):
## Code Review

Great first test suite! Here's my full review:

### ✅ What's solid

**`tests/test_transcribe.py` (22 tests)**
- Good coverage of `parse_vtt` edge cases: HTML tag stripping, rolling/identical deduplication, comma decimal separators, and empty files.
- - The `_vtt()` helper keeps fixtures concise.
- - - `TestDedupe` and `TestFilterRange` test the pure functions cleanly in isolation.
- - - - `TestFormatTranscript` roundtrip via `format_time` → `parse_time` is a nice sanity check.
**`tests/test_frames.py` (25 tests)**
- `TestParseTime` covers all timestamp formats (`MM:SS`, `HH:MM:SS`, decimal variants) and passthrough of numeric types.
- - `TestAutoFpsFocus.test_denser_than_auto_fps_for_short_clips` is a good behavioral assertion (not just a value check).
- - - `fail-fast: false` in the CI matrix is the right call — you want to see all Python version failures at once.
**`tests/test_download.py` (28 tests)**
- `TestIsUrl` covers platform-specific paths (Windows backslash path, tilde, bare filename) — these are the cases most likely to regress.
- - `TestPickSubtitle` tests the `en` preference logic and the `en-US`/`en-GB` variant handling correctly.
**`scripts/download.py` & `scripts/frames.py` bug fixes**
- Platform-neutral install hints are correct and consistent across both files.
- - The `auto_fps_focus` docstring fix (em-dash → hyphen) is a minor but appreciated cleanup.
---

### ⚠️ Suggestions / things to watch

1. **`test_unknown_extension_warns_but_proceeds` may be fragile.** The test asserts `"warning" in captured.err.lower() or "xyz" in captured.err`. If `resolve_local` doesn't currently print to stderr for unknown extensions, this test will silently pass on the `or` branch without actually validating the warning. Consider splitting into two separate assertions or using `pytest.warns` if a warning is raised.
2. **`test_tilde_expansion` doesn't actually test tilde expansion.** The comment says "Monkeypatch Path.expanduser to return tmp_path / 'video.mp4'" but the test just passes a real absolute path. It would be more valuable to call `resolve_local("~/video.mp4")` with a monkeypatched `expanduser`, or remove the test and rename it to `test_absolute_path_works`.
3. **`TestAutoFps.test_30_second_boundary` uses a hard-coded expected max target of 30.** If the fps budget algorithm changes, this test breaks without a clear explanation of why 30 is expected. Consider adding a brief comment explaining the intent (e.g., "30s clip should get at most 1 fps").
4. **CI workflow only runs on `ubuntu-latest`.** The bug fixes are specifically about macOS/Windows install hints. It might be worth adding a `windows-latest` runner row to actually exercise the Windows code path — even if just for the platform-neutral error message test.
5. **No `conftest.py` or `__init__.py` in `tests/`.** This is fine for now with `pytest`, but worth noting for future contributors. A `conftest.py` could house the `_vtt()` helper if more test files need it.
---

### Summary

This is a clean, well-structured contribution. The tests are readable, isolated, and don't require any external binaries — exactly right for a CI-first setup. The bug fixes are correct. The suggestions above are minor; none are blockers. Ready to merge once CI passes.

Issue #2, opened 2026-04-30
Hardening: insert `--` before URL in yt-dlp argv to block option injection


dayanimo:
Hi! Found this while reviewing the skill for personal use. Just a heads-up about a small argv-hygiene gap, plus a one-line fix.

## What

In `scripts/download.py`, the user-supplied URL is passed as the last positional argument to `yt-dlp` without a `--` separator:

https://github.com/bradautomates/claude-video/blob/main/scripts/download.py#L67-L82

```python
cmd = [
    "yt-dlp",
    "-N", "8",
    ...
    "-o", output_template,
    url,           # ← no `--` before this
]
subprocess.run(cmd, ...)
```

`subprocess` itself is correctly invoked with an argv array (no shell injection — nice), but `yt-dlp`'s own argument parser will treat any string starting with `-` / `--` as a flag rather than a URL. yt-dlp ships several flags that load attacker-controlled state from disk or run code (`--config-location`, `--exec`, `--ffmpeg-location`, `--paths`, `--cookies`, …).

## Practical risk

In this repo it's mostly mitigated by `is_url` filtering inputs that don't have an `http(s)` scheme, so a bare `--exec=...` falls through to `resolve_local` and errors out. But:

- `is_url` only requires `urlparse(source).scheme in ("http", "https")` — it doesn't validate the rest of the URL shape.
- The `/watch` slash command flows user-typed input straight into this script as the source.
- A user who is socially engineered into pasting a "video link" with a leading `--` (or a craftily-shaped `http://…` that yt-dlp's parser misreads) gets argv-injection into `yt-dlp`, which can pivot to RCE under their account.

It's a low-probability attack today, but the fix is one token and makes it impossible by construction.

## Fix

Two cheap changes, either is fine — applying both is best:

1. **Insert `\"--\"` immediately before `url`** so yt-dlp stops option parsing first:

   ```python
   cmd = [
       "yt-dlp",
       "-N", "8",
       ...
       "-o", output_template,
       "--",
       url,
   ]
   ```

2. **Tighten `is_url`** to also reject `source.startswith(\"-\")` and inputs without a non-empty `netloc`. Belt-and-suspenders against future regressions.

Same hardening would also be nice on the ffmpeg/ffprobe argv lists in `scripts/frames.py` and `scripts/whisper.py` — currently safe because `Path.resolve()` produces absolute paths starting with `/`, but the uniform treatment is cheaper than reasoning about it.

## Why I'm filing

[...]

PR #5, opened 2026-05-01
feat: add local whisper-cli transcription backend

Zaargh:
## Summary

- Adds local `whisper-cli` (whisper.cpp) as a third transcription backend, preferred over Groq/OpenAI API when available
- Auto-detects binary via PATH and GGML models in `~/.cache/whisper/` or brew cellar
- Extracts audio as 16kHz mono WAV (native format for whisper.cpp) instead of mp3
- New CLI flag `--whisper local` to force local backend
- Config: `WHISPER_MODEL` env var for explicit model path, `WHISPER_LOCAL=false` to disable
- Updates setup.py preflight, session hook, and installer to detect and report local whisper

## Motivation

Users with `whisper-cpp` installed locally can transcribe without API keys, network access, or cost. Local transcription is also more private since audio never leaves the machine.


Zaargh:
Rebased onto `main` — conflicts in `CHANGELOG.md` (kept the new 0.1.3 release block alongside the Unreleased entry) and `scripts/whisper.py` (kept `out_path.resolve()` from the subprocess hardening in #2, dropped the redundant `-b:a 64k` since codec args are now branched on wav vs mp3).

PR is `MERGEABLE` / `CLEAN` again. Friendly bump in case it slipped past — happy to address any review feedback.

Zaargh:
Closing this one in favor of #41.

PR #18, opened 2026-05-09
Add local Whisper backend (mlx_whisper / openai-whisper)

glen-carlson:
## Summary

Adds a third Whisper backend that shells to a local binary instead of POSTing audio to Groq or OpenAI. Useful for:

- **Offline use** — `/watch` works on a plane / behind a corporate proxy.
- **Privacy** — audio never leaves the machine.
- **The 25 MB cap** — local has no upload limit, so long-form videos with no captions stop being a problem.
- **Apple Silicon speed** — `mlx_whisper` runs `whisper-large-v3-turbo` faster than the cloud round-trip on an M-series machine, with no per-call cost.

The implementation reuses the existing audio-extraction step (mono 16 kHz mp3) and shells to a binary that takes the openai-whisper CLI shape:

```
<bin> <audio> --output-format json --output-dir <dir> [--model <id>]
```

`mlx_whisper` and `openai-whisper` both expose this exact CLI, so one code path handles both. The resulting `<stem>.json` matches the Whisper API `verbose_json` schema, so `_segments_from_response` works unchanged.

## How to use

```bash
/watch ~/Movies/clip.mov                  # auto-detected if no cloud key set
/watch <url> --whisper local              # force the local backend
WATCH_WHISPER_BACKEND=local /watch <url>  # global override
```

Optional knobs:

- `WATCH_LOCAL_WHISPER_BIN=/path/to/binary` — point at any compatible binary.
- `WATCH_LOCAL_WHISPER_MODEL=large-v3` — override the model id.

## Resolution order

When `--whisper` isn't set, the priority is:

1. `WATCH_WHISPER_BACKEND` env / `.env` (groq | openai | local)
2. `GROQ_API_KEY`
3. `OPENAI_API_KEY`
4. local binary (`mlx_whisper` → `openai-whisper`)

So existing users see **zero behaviour change** — cloud keys still win when present. Local only kicks in when nothing else is configured, or when explicitly requested.

## Changes

- `scripts/whisper.py` — new `resolve_backend()` helper, `_post_local()` runner, `_resolve_local_bin()` probe. `load_api_key()` retained unchanged for back-compat with anyone importing it.
- `scripts/watch.py` — `--whisper` accepts `local`; uses `resolve_backend`; targeted hint when `--whisper local` fires without a binary on PATH.
- `scripts/setup.py` — `_have_local_whisper()` probe; `--check` / `--json` / installer report the local backend; setup is "ready" when either a cloud key or a local binary is present. `local_whisper_bin` added to the `--json` snapshot.
- `hooks/scripts/check-setup.sh` — SessionStart hook is silent when local Whisper is available, even with no cloud key.
- `README.md` / `SKILL.md` / `CHANGELOG.md` — documentation.

## Test plan

- [x] `python3 scripts/whisper.py /tmp/test.mp3 --backend local` end-to-end on Apple Silicon with `mlx-community/whisper-large-v3-turbo` — JSON parses, segments normalise to `{start, end, text}`.
- [x] `resolve_backend(None)` returns `("local", "/opt/homebrew/bin/mlx_whisper")` when no cloud key is set.
- [x] `resolve_backend("groq")` returns `(None, None)` when GROQ_API_KEY is unset (existing behaviour preserved).
- [x] `setup.py --check` exits 0 with no output when local binary is present and no cloud key is set.
- [x] `hooks/scripts/check-setup.sh` exits 0 with no output in the same situation (after first run).
- [ ] CI: I haven't run the existing test suite — there isn't one in the repo. Happy to add unit tests for `resolve_backend` if you'd like.

## Notes for review

- I kept `load_api_key()` exported and unchanged so external callers don't break. New code uses `resolve_backend()`.
- The `WATCH_LOCAL_WHISPER_BIN` env knob accepts either a name on PATH or an absolute path, mirroring how `WATCH_LOCAL_WHISPER_MODEL` works.
- I followed the existing "pure stdlib, no SDK" convention — no new pip deps, no new imports beyond what was already there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to `/watch` are documented here.
 
+## [Unreleased]
+
+### Added
+- Local Whisper backend. `--whisper local` (or `WATCH_WHISPER_BACKEND=local`) shells to a local Whisper binary instead of POSTing to Groq/OpenAI — no API key needed and audio never leaves the machine. Auto-detects `mlx_whisper` (Apple Silicon) and `openai-whisper`, falling back to either when no cloud key is set. Override the binary with `WATCH_LOCAL_WHISPER_BIN` and the model with `WATCH_LOCAL_WHISPER_MODEL`.
+- `setup.py --json` and the SessionStart hook now report local Whisper availability; setup is considered "ready" when either a cloud key or a local binary is present.
+
+### Changed
+- `whisper.py` exports a new `resolve_backend(preferred)` helper that handles all three backends. The original `load_api_key` is preserved for backward compatibility but only knows about the cloud backends.
+
 ## [0.1.3] — 2026-05-09
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ On the first `/watch` call, the skill runs `scripts/setup.py --check`. If `ffmpe
 - **Linux** — prints the exact `apt` / `dnf` / `pipx` commands.
 - **Windows** — prints the `winget` / `pip` commands.
 - **API key** — scaffolds `~/.config/watch/.env` (mode `0600`) with commented placeholders for `GROQ_API_KEY` (preferred) and `OPENAI_API_KEY`.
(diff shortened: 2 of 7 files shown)


glen-carlson:
Clearing my old open proposals; reopen if this is still wanted.

PR #35, opened 2026-06-09
Add Deepgram backend: transcribe audio over Whisper's 25 MB limit (45 min+ videos)

drlee91:
## Problem

Groq and OpenAI Whisper both reject uploads over **25 MB** with HTTP 413. At the 64 kbps mono extraction rate that's roughly **45 minutes of audio** — so any longer caption-less video currently comes back frames-only, with no transcript possible at all:

```
[watch] audio: 27625 kB — uploading to groq Whisper…
[watch] whisper fallback failed: Whisper request failed: HTTP Error 413: Payload Too Large
       — {"error":{"message":"Request Entity Too Large",...,"code":"request_too_large"}}
```

This bites in practice: long tutorials, podcasts, talks, VODs — and YouTube caption pulls can also fail with 429 rate limits, which makes the audio fallback the only path for those videos.

## Approach

Add **Deepgram's pre-recorded API** (`nova-2`) as a third backend. Its file limit is ~2 GB, so the 25 MB cliff disappears. Routing keeps current behavior unchanged for existing users:

- **Audio ≤ 24 MB:** exactly as before — Groq preferred, OpenAI fallback. Deepgram additionally catches the case where Whisper errors out.
- **Audio > 24 MB:** routed straight to Deepgram instead of failing with 413. Without a Deepgram key, Whisper is still attempted so the user sees the clear 413 error.
- **Standalone:** if `DEEPGRAM_API_KEY` is the only configured key, Deepgram serves as the primary backend.
- **Override:** `--whisper deepgram` forces it.

### Design notes

- **Pure stdlib**, matching `whisper.py` — `urllib` only, no SDK dependency. Same `{start, end, text}` segment shape (from Deepgram `utterances`, with paragraph/transcript fallbacks), same `SystemExit` error contract, same key resolution (env → `~/.config/watch/.env` → `./.env`).
- `detect_language=true` keeps it language-agnostic like the Whisper path.
- The module is named `deepgram_backend` (not `deepgram`) so it can never shadow or be shadowed by the real Deepgram SDK.
- `setup.py` accepts `DEEPGRAM_API_KEY` as a valid transcription key, scaffolds a commented placeholder, and includes it in the install hints. `--check`/`--json` semantics are unchanged otherwise.
- Considered chunked uploads to stay within Whisper's 25 MB instead, but that needs segment-boundary handling, per-chunk timestamp offsetting, and N sequential uploads — a second provider that follows the existing backend contract is simpler and also gives users a provider choice. Happy to adjust if you'd prefer the chunking route.

## Verification

- **59-min YouTube video** (caption pull 429'd, audio 27 MB): routed to Deepgram, returned **628 timestamped segments**, full report, exit 0. Previously: frames-only.
- **Small file regression:** 40 s local clip without flags → still transcribed via **Groq** (unchanged default).
- **Forced:** `--whisper deepgram` with a Groq key present → Deepgram used.
- **Standalone:** only `DEEPGRAM_API_KEY` configured → Deepgram auto-selected as primary; `setup.py --json` reports `ready`.
- **No keys at all:** updated guidance message lists all three options; `--check` exits 3 as before.

Docs (README + SKILL.md) updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Claude is great at reading and synthesizing — but until now, video was the one
 1. **You paste a video and a question.** URL (anything yt-dlp supports — YouTube, Loom, TikTok, X, Instagram, plus a few hundred more) or a local path (`.mp4`, `.mov`, `.mkv`, `.webm`).
 2. **`yt-dlp` downloads it.** For URLs, into a temp working directory. For local files, no download — just probed in place.
 3. **`ffmpeg` extracts frames at an auto-scaled rate.** The frame budget is duration-aware: ≤30s gets ~30 frames, 30-60s gets ~40, 1-3min gets ~60, 3-10min gets ~80, longer gets 100 sparsely. Hard ceilings: 2 fps, 100 frames. JPEGs at 512px wide by default — bump with `--resolution 1024` if Claude needs to read on-screen text.
-4. **The transcript comes from one of two places.** First try: `yt-dlp` pulls native captions (manual or auto-generated) from the source. Free, instant, accurate-ish. Fallback: extract a mono 16 kHz audio clip and ship it to Whisper — Groq's `whisper-large-v3` (preferred — cheaper and faster) or OpenAI's `whisper-1`.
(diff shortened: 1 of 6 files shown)


drlee91:
Closing — the underlying problem (Groq/OpenAI Whisper's 25 MB upload limit failing on long videos) is now solved differently in `304b639b` (Whisper auto-chunking, part of 0.2.0). That's the simpler fix and doesn't need a new API key, so I don't think it's worth carrying Deepgram as a second backend for the same problem.

Thanks for the auto-chunking work — closing this in favor of it.

PR #40, opened 2026-06-30
feat(watch): high-density mode for frame-by-frame / transition analysis

christianlappin:
## What & why

Auto-mode samples at most **2 fps**, so a cut, fade, wipe, or fast on-screen change shorter than ~500ms can fall between frames. There was no way to ask for denser sampling — `--fps` was silently clamped to 2, and `--max-frames` was hard-capped at 100. This adds an opt-in path to sample at the source's **native rate with no skipped frames**, for frame-by-frame / transition analysis.

All default behavior (and its token cost) is unchanged — these are opt-in overrides.

## Changes

- **`--every-frame`** — samples at the video's native fps (every decoded frame, no gaps), capped by `--max-frames`. `get_metadata` now reports native fps parsed from `avg_frame_rate` / `r_frame_rate`.
- **`--fps` honored verbatim** — no longer clamped to 2 fps, so callers can request native-rate sampling for fast transitions. The 2 fps ceiling now applies to **auto-mode only**.
- **`--max-frames` ceiling 100 → 1000** (`HARD_MAX_FRAMES`) — a runaway guard rather than a working limit. Auto-mode duration budgets are unchanged.
- **Guardrails** — `watch.py` prints a token-cost warning above ~150 frames, and a hint when high-density mode runs over a full long video (where `--max-frames` would silently truncate). Docs steer users to pair high-density with a tight `--start`/`--end`.
- Docs: new "High-density mode" sections in `SKILL.md` and `README.md`, `CHANGELOG.md` entry.

## Usage

```bash
# Every frame across a 3s transition (3s × 30fps = 90 frames, so allow >= 90)
/watch "$URL" --start 0:45 --end 0:48 --every-frame --max-frames 120

# Or pin an exact rate (now honored, not clamped)
/watch "$URL" --start 0:45 --end 0:48 --fps 30 --max-frames 120
```

## Testing

Smoke-tested against a generated 3s/30fps `testsrc` clip:
- auto-mode → 2 fps (unchanged)
- `--every-frame` → 30 fps, native rate; truncates to `--max-frames` as documented
- `--fps 10` → honored verbatim (not clamped to 2)
- `--every-frame --max-frames 20` → capped at 20 frames

`python3 -m py_compile` passes on both scripts.

## Model Used

- [x] I verified the changes in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to `/watch` are documented here.
 
+## [Unreleased]
+
+### Added
+- **High-density mode for frame-by-frame / transition analysis.** New `--every-frame` flag samples at the video's native fps (no skipped frames), capped by `--max-frames`. It bypasses the detail engines and perceptual dedup so no frame is ever dropped. Best paired with a tight `--start`/`--end` window.
+
+### Changed
+- `--fps` is now honored verbatim instead of being silently clamped to 2 fps, so callers can request native-rate sampling to catch fast cuts/fades. The 2 fps ceiling now applies to auto-mode only.
+- `--max-frames` hard ceiling raised from 100 to 1000 (`HARD_MAX_FRAMES`) as a runaway-guard rather than a working limit. Auto-mode budgets are unchanged, so default scans cost the same.
+- `get_metadata` now reports the source's native `fps` (parsed from `avg_frame_rate`/`r_frame_rate`).
+- `watch.py` prints a token-cost warning above ~150 frames and a hint when high-density mode runs over a full long video (where `--max-frames` would truncate).
+
 ## [0.2.0] — 2026-06-29
 
 ### Added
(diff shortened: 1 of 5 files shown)


rickgonzalez:
A narrated explainer of this PR - I hope it is helpful. https://www.lenzon.ai/viewer/cmrb2qjdi000cpyxah78yv2xc?voice=google-chirp3

christianlappin:
Rebased onto current main (the 0.2.0 `skills/watch` package layout) — conflicts resolved, now MERGEABLE.

The rebase was structural, so a couple of integration decisions worth flagging:

- **`--every-frame` bypasses the detail engines and the perceptual dedup pass.** Dedup deliberately drops near-identical frames, which would silently break this flag's contract ("no skipped frames"), so high-density runs go through the uniform extractor at native fps. `--timestamps` cue frames still merge on top.
- **`--fps` verbatim** replaces the clamp in the new architecture too — the 2 fps ceiling is auto-mode only, exactly as before the restructure.
- Report line shows an `every-frame` engine and the detail line reads `every-frame (high-density)` so it's obvious the dial was bypassed.

All 71 tests green, plus a smoke test: a 1s window of a 30fps clip yields exactly 30 frames, none dropped.

PR #41, opened 2026-06-30
feat(watch): local Whisper backend (whisper.cpp / mlx), no API key needed

christianlappin:
## What & why

Today `/watch` requires a Groq or OpenAI key the moment a video lacks captions (local files, most TikToks, some Vimeos). This adds an **on-device** transcription path: free, private, no network, no 25 MB upload cap, and faster than a cloud round-trip on Apple Silicon.

Backend priority becomes **local → Groq → OpenAI**, first-available-wins, so existing key-only setups behave exactly as before.

There are three open local-Whisper PRs ([#5](https://github.com/bradautomates/claude-video/pull/5), [#18](https://github.com/bradautomates/claude-video/pull/18), [#32](https://github.com/bradautomates/claude-video/pull/32)); this one auto-detects **both** engine families rather than picking one, and corrects a factual issue in #32 (Homebrew's `whisper-cpp` no longer ships a model — see note below).

## Changes

- **whisper.cpp** (`whisper-cli` / `whisper-cpp` / `main` binary + a `ggml-*.bin` model). Metal-accelerated on Apple Silicon. Auto-detected from `/opt/homebrew/share/whisper-cpp/`, `~/.cache/whisper/`, or `$WHISPER_MODEL`, preferring quantized `large-v3-turbo`.
- **mlx_whisper / openai-whisper** Python CLIs — detected on PATH; they fetch their own model. Flag-spelling differences (hyphen vs underscore) are handled.
- `extract_audio` emits **16 kHz PCM WAV** for `.wav` targets (whisper.cpp's native input); the cloud path still uses mp3.
- `setup.py --check` treats a usable local engine as a valid backend, so no API key is demanded when local Whisper is installed.
- `--whisper` now accepts `local`. Pure stdlib preserved — local engines are external binaries, like `ffmpeg`/`yt-dlp`.
- Docs updated (SKILL.md, README.md, CHANGELOG.md).

> **Note for macOS users:** `brew install whisper-cpp` installs the binary but **not** a model. Download one separately, e.g.
> `curl -fL -o /opt/homebrew/share/whisper-cpp/ggml-large-v3-turbo-q5_0.bin https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-large-v3-turbo-q5_0.bin` (~547 MB).

## Testing

On Apple M4 (24 GB), whisper-cpp + `ggml-large-v3-turbo-q5_0`:
- `whisper.py <video> --backend local` → extracts WAV, runs `whisper-cli`, returns `{start,end,text}` segments with correct timestamps.
- `watch.py <local-file>` with **no API key set** → auto-selects local, transcript reported `via whisper (local)`.
- `setup.py --check` → exit 0; `--json` → `status=ready, whisper_backend=local`.
- `python3 -m py_compile` passes on all scripts.

## Model Used

- [x] I verified the changes in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 All notable changes to `/watch` are documented here.
 
+## [Unreleased]
+
+### Added
+- **Local Whisper backend — no API key, no network, no upload limit.** Auto-detects two engines: `whisper.cpp` (`whisper-cli` binary + a `ggml-*.bin` model, Metal-accelerated on Apple Silicon) and the `mlx_whisper` / `whisper` Python CLIs. `--whisper local` forces it. whisper.cpp gets 16 kHz WAV; the Python CLIs take the existing mp3.
+- `--accurate` flag (use `large-v3`) and `--model turbo|large-v3|<path>` for picking the local whisper.cpp model by friendly alias — no more typing a full `WHISPER_MODEL=/…/ggml-*.bin` path. `$WHISPER_MODEL` still works for an explicit override.
+- Model auto-pick now prefers full `large-v3-turbo` over the quantized variant when both are present.
+- `WATCH_DISABLE_LOCAL_WHISPER=1` skips local-engine detection entirely (forces cloud backends; also keeps the setup tests hermetic on machines with whisper installed).
+
+### Changed
+- Whisper backend priority is now **local → Groq → OpenAI** (was Groq → OpenAI). The first available backend wins; key-only setups are unaffected. `--whisper` now accepts `local`.
+- `setup.py --check` treats a usable local engine as a valid backend, so no API key is required when local Whisper is installed.
+- `extract_audio` emits 16 kHz PCM WAV when the output path ends in `.wav` (whisper.cpp's native input), else the existing mp3.
+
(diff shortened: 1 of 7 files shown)


christianlappin:
Rebased onto current main — conflicts resolved, now MERGEABLE. Squashed my two commits into one in the process.

Integration notes:

- The local backend now sits alongside 0.2.0's cloud auto-chunking: local has no upload cap, so the chunk/split path applies to Groq/OpenAI uploads only. `extract_audio`'s WAV branch merged cleanly with the chunking helpers.
- **Heads-up on `tests/test_setup.py`:** with local now winning the backend priority, three of the new setup tests fail on any dev machine that has whisper installed (they assert the cloud-key paths, and detection finds the real local install). I added a `WATCH_DISABLE_LOCAL_WHISPER=1` env kill-switch honored by both `setup.py` and `whisper.py`, and set it in the test helper — same pattern the helper already uses to scrub real `GROQ_API_KEY`/`OPENAI_API_KEY` from the developer's shell. CI machines without whisper were never affected.

All 71 tests green locally (on a machine that *does* have whisper.cpp + large-v3-turbo installed, which is exactly the case the kill-switch covers).

PR #45, opened 2026-07-04
fix: secrets-file permission check on native Windows (NTFS)

hekwweiss89:
## Problem

On native Windows (Git Bash / MSYS), the SessionStart hook's permission check on `~/.config/watch/.env` warns forever:

```
/watch: WARNING — ~/.config/watch/.env has permissions 644 (should be 600).
```

MSYS synthesizes Unix permission bits on NTFS — non-executable files always report `644` from `stat`, and `chmod 600` is a silent no-op. So the suggested fix can never satisfy the check, and Windows users see the warning on every session start even after locking the file down with real NTFS ACLs.

## Fix

Platform-split the check:

- **MINGW/MSYS/CYGWIN:** inspect the actual ACL via `icacls` and warn only when a broad-access principal holds an entry. Detection is **SID-based** (`S-1-1-0` Everyone, `S-1-5-32-545` BUILTIN\Users, `S-1-5-11` Authenticated Users) so it works on any Windows display language — group names are localized (e.g. "Todos"/"Usuarios" on Spanish Windows), SIDs are not. The suggested fix line now uses `icacls` instead of `chmod`.
- **Unix/macOS:** unchanged (original `stat` check).

Fails open: if `icacls`/`cygpath` are unavailable, the check is skipped rather than misfiring.

## Testing

On Spanish-locale Windows 11, Git Bash, against the live config file:

- clean ACL (owner + SYSTEM + Administrators only) → hook silent
- `icacls <file> /grant *S-1-5-32-545:R` (BUILTIN\Users via SID) → warning fires, names the SID
- `icacls <file> /remove *S-1-5-32-545` → silent again
- `bash -n` clean; Unix branch logic untouched


willrenzo39:
**Independent verification on a second machine — approach confirmed, plus one bug found (one-token fix)**

Setup: Windows 11 Pro (build 26200), Git for Windows bash, `C:` mounted `noacl`, English-locale system, AzureAD account. I hit the same permanent 644 warning this PR addresses and had written an equivalent name-based `icacls` check locally before finding this PR — the SID-based detection here is strictly better (group names are localized, SIDs aren't), so I tested this PR's logic on my machine instead of opening a duplicate.

**Root cause confirmed.** `stat -c '%a'` reports a fixed synthetic `644` on the noacl mount, `chmod 600` is a no-op, and the real NTFS ACL governs access (my `.env` carries a single owner ACE and is tighter than 600 while still reporting 644).

**Detection approach verified.** I extracted the MINGW/MSYS/CYGWIN branch verbatim (with `CONFIG_FILE` parameterized) and granted a scratch file `BUILTIN\Users:R` via `icacls /grant *S-1-5-32-545:R`. The raw `/findsid` call finds it and echoes the path exactly as passed (`SID Found: <path>.`), and `MSYS_NO_PATHCONV=1` correctly stops the `/findsid` flag from being mangled.

**Bug: `grep -qiF` aborts under the C locale on Git for Windows' grep 3.0, so the check silently never fires in the exact environment SessionStart hooks run in.**

Git for Windows still ships GNU grep 3.0, whose case-insensitive fixed-string matcher crashes when the locale is C/unset:

```
$ LANG= LC_ALL= sh -c 'echo hello | grep -qiF hello'
Aborted   (exit 134 = SIGABRT)
```

Interactive shells usually have a UTF-8 locale (where `-iF` works), which is probably why testing passed — but hooks are spawned with empty `LANG`/`LC_*`. Reproduced under Claude Code's non-interactive shell on my machine: each `/findsid | grep -qiF` pipeline dies, the `if` reads it as "no match", `broad_sid` stays empty, and the hook stays silent on a file that is actually granted to `BUILTIN\Users` — a false negative that defeats the check in its primary environment. Trigger matrix on grep 3.0:

| invocation | C / unset locale | en_US.UTF-8 |
|---|---|---|
| `grep -iF` | **abort (134)** | ok |
| `LC_ALL=C grep -iF` | **abort (134)** | — |
| `grep -F` | ok | ok |
| `grep -iE`, `grep -i` (BRE) | ok | ok |

**Suggested fix — drop the `-i`:**

```bash
if MSYS_NO_PATHCONV=1 icacls "$win_path" /findsid "*$sid" 2>/dev/null | grep -qF "$win_path"; then
```

Case-folding buys nothing here: `icacls` echoes the path verbatim as passed, so an exact `-F` match is safe. (Note `LC_ALL=C` does *not* work around the abort on this grep build, and `-iE` would treat the backslashes in `$win_path` as regex escapes — dropping `-i` is the clean fix.)

Retested with only that change, same machine, C locale:

- scratch file with `BUILTIN\Users:(R)` → `WARNING — … ACL grants broad access (SID S-1-5-32-545)` ✓
- real `~/.config/watch/.env` (single owner ACE) → silent ✓
- `bash -n` clean ✓

With that one token removed, this fixes the Windows false positive without introducing a false negative. Would love to see it merged — I'm running the equivalent patch locally in the meantime.

PR #53, opened 2026-07-06
Auto-write TRANSCRIPT.md and fix rolling-caption overlap in dedupe

adamharveyai:
## Summary

Two related changes to transcript handling:

1. **Fix rolling-caption overlap in `transcribe._dedupe`.** YouTube's default rolling display puts two lines per cue (previous + current), and consecutive cues overlap by the previous "current" line. The old dedupe only handled identical + strict extension cases, so a lot of duplication survived into the printed report — e.g.:

   ```
   [00:04] model and the most important thing that
   [00:06] model and the most important thing that you're going to build this year is an
   [00:08] you're going to build this year is an agentic operating system. And in this
   ```

   New `_strip_overlap` helper finds the longest prev-suffix / curr-prefix word match and drops it, so each segment carries only its new content.

2. **Auto-write `TRANSCRIPT.md` to the work dir on every run.** New `transcribe.write_transcript_file(segments, out_path, ...)` writes a paragraph-grouped Markdown transcript (~30s buckets) so the file reads as prose rather than a wall of two-second cues. `watch.py` calls it once the transcript is finalized (captions or Whisper) and prints the path in the report so the agent (and user) can find it.

Motivation: on a fresh install I ran `/watch` on a 19-minute YouTube video and the report's transcript section had heavy duplication (610 raw cues, half of them overlapping their neighbour). Also, having only the raw `.vtt` in the work dir meant I had to hand-write a small script to dedupe + paragraph-group it into a readable file — felt like something the skill should just do.

## Test plan

- [x] On a 19-min YouTube video with captions: 610 raw cues collapse to 306 clean segments; report transcript reads as continuous prose; `TRANSCRIPT.md` renders 38 readable paragraphs with no duplication.
- [ ] On a video without captions where Whisper is used (Whisper segments have no overlap pattern, so the dedupe change is a no-op there and `TRANSCRIPT.md` should still write cleanly).
- [ ] On a focused range (`--start` / `--end`) — the header now includes a `Focus range: ...` line and only in-range segments are written.

## Notes

- Bucket size is a keyword-only arg on `write_transcript_file` (`bucket_seconds=30`) in case anyone wants finer/coarser paragraphs later.
- `transcript_file` is `None`-safe: if the write fails for any reason, `watch.py` swallows the exception, prints a stderr note, and the rest of the report still prints.
- No new dependencies.


eragonlonelyboy-lab:
Independent verification of this PR against a real-world fixture — it holds up, and the numbers are worth having on the thread.

**Fixture:** a 19:21 YouTube screencast (`8rVQuZlRaqo`) with auto-generated English captions, pulled by yt-dlp as `video.en.vtt`. This is exactly the rolling-caption pattern @adamharveyai describes, where each cue repeats the tail of the previous one and appends new content:

```
[17:11] at what I've done and be like, "Huh, we could make that better." Um and that was
[17:12] could make that better." Um and that was sort of sort of the idea. I'm sure there
[17:14] sort of sort of the idea. I'm sure there are many ways you can make this better,
```

Neither existing branch in `_dedupe` catches this. It is not an exact duplicate, and it is not a prefix extension (`seg["text"].startswith(prev + " ")` is false because the new cue *starts* mid-way through the previous one). So it falls through to `out.append(seg)` untouched.

**Measured on `main` vs this branch:**

| | segments | words |
|---|---|---|
| current `main` | 636 | 9,047 |
| this PR | 319 | **4,532** |

**50% of the transcript on `main` is redundant text.** For a 19-minute video that is roughly 4,500 wasted words in the agent's context on every run — and it scales linearly with video length, so it is worst exactly where transcripts already hurt most.

**Correctness check.** I looked for over-stripping, since a greedy word-overlap match could plausibly eat legitimate repeated speech. Scanning the deduped output for repeated 6-grams returns 7 hits, and every one is genuine speaker repetition, not an artifact:

```
x6  "and so on and so forth."
x3  "I've given it context as to"
x2  "And to make a long story"
```

Those are real — the speaker actually says them. No legitimate content was lost.

I also independently wrote a character-level version of the same overlap strip before finding this PR, and it converges on the identical output (4,532 words). The word-level implementation here is the better choice: it cannot split a token mid-word, which a character-level match can do on repeated substrings.

**One conservative gap, noted rather than requested.** `_strip_overlap` compares whole words, so punctuation blocks a match that a human would call an overlap:

```
prev: "... and so on and so forth"
curr: "so on and so forth. Okay, if you like ..."
```

`"forth"` vs `"forth."` fails the comparison, and the overlap goes unstripped. This fails *safe* — it under-strips rather than corrupting text — and normalising punctuation would raise the risk of eating real repetition. I would leave it as is. Flagging it only so the behaviour is documented rather than discovered later.

Nice catch on this one, @adamharveyai — the three-pattern breakdown in the docstring is a much clearer statement of the problem than the original comment, and the `TRANSCRIPT.md` writer in the same PR is a good companion change. Would be good to see this land.

*(Environment: Windows 11, Python 3.14, ffmpeg 8.1.2, yt-dlp 2026.07.04.)*

PR #72, opened 2026-07-13
Add WHISPER_BASE_URL for self-hosted / local Whisper servers

alexschar:
## What

Adds `WHISPER_BASE_URL` so `watch` can use **any OpenAI-compatible** `/v1/audio/transcriptions` endpoint instead of Groq/OpenAI — a local [whisper.cpp](https://github.com/ggml-org/whisper.cpp) `whisper-server`, a self-hosted box, or a proxy.

```bash
# ~/.config/watch/.env
WHISPER_BASE_URL=http://127.0.0.1:8178/v1
```

Audio then never leaves the machine, and there's no API key, no rate limit, and no per-minute cost.

## Why

The Whisper fallback is currently reachable only via Groq or OpenAI. For users who care about privacy (recordings of meetings, interviews, anything confidential), work offline, or want to transcribe large volumes without cost or rate limits, a self-hosted server is the natural answer — and since `watch` already speaks the OpenAI multipart API and already requests `response_format=verbose_json`, it turns out the *only* thing standing in the way is the hardcoded hostname.

whisper.cpp's bundled `whisper-server` can serve the OpenAI route directly, so this needs no glue code at all:

```bash
brew install whisper-cpp
whisper-server -m ggml-large-v3-turbo.bin \
  --host 127.0.0.1 --port 8178 \
  --inference-path /v1/audio/transcriptions --convert
```

## Changes

- **`WHISPER_BASE_URL`**, plus optional `WHISPER_MODEL` and `WHISPER_API_KEY`, read from the environment *or* `~/.config/watch/.env` — the same two places the existing keys come from. (`_from_dotenv` is promoted to module scope so a `read_setting()` helper can share it; `load_api_key` keeps its exact behavior.)
- **No API key required when `WHISPER_BASE_URL` is set.** Most self-hosted servers need no auth, and `load_api_key()` previously hard-exited without a cloud key, which would have blocked the whole path.
- **Honest progress output.** With a custom base URL, `watch` now prints `local` (or the hostname) rather than `groq`/`openai` for a request that never goes there. People choose a custom base URL precisely to keep audio off third-party servers — reporting "uploading to groq" while hitting localhost would be actively misleading.
- README documents the setup.

## Backwards compatibility

Unset `WHISPER_BASE_URL` and **nothing changes** — same endpoints, same models, same messages. Tests cover this explicitly.

## Testing

13 new tests in `tests/test_whisper_base_url.py` (endpoint resolution, dotenv vs env precedence, key short-circuit, backend labeling). **Full suite: 84 passed.**

Verified end-to-end against a real whisper.cpp `whisper-server` on Apple Silicon (M4): a video with no captions transcribes correctly through the local server, with no `GROQ_API_KEY`/`OPENAI_API_KEY` present anywhere in the environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,36 @@ Captions cover the majority of public videos for free. The Whisper fallback only
 | Download + native captions | `yt-dlp` + `ffmpeg` | Free |
 | Whisper fallback (preferred) | [Groq API key](https://console.groq.com/keys) — `whisper-large-v3` | Cheap, fast |
 | Whisper fallback (alt) | [OpenAI API key](https://platform.openai.com/api-keys) — `whisper-1` | Standard pricing |
+| Whisper fallback (self-hosted) | `WHISPER_BASE_URL` → any OpenAI-compatible server | Free, private, offline |
 | Disable Whisper entirely | `--no-whisper` | Free, frames-only when no captions |
+
+### Self-hosted / local Whisper
+
+Set `WHISPER_BASE_URL` (in the environment or `~/.config/watch/.env`) to point `watch` at
+any OpenAI-compatible `/v1/audio/transcriptions` endpoint instead of a cloud provider —
+audio then never leaves your machine, and there is no API key, rate limit, or per-minute cost.
+
+```bash
+# ~/.config/watch/.env
+WHISPER_BASE_URL=http://127.0.0.1:8178/v1
+```
+
+No API key is required when `WHISPER_BASE_URL` is set (most self-hosted servers need no
+auth). If yours does, set `WHISPER_API_KEY`. To send a specific model name, set
+`WHISPER_MODEL`. Progress output reports the actual host (`local`, or the hostname) rather
+than naming a cloud provider it isn't using.
+
+A local server that works out of the box is [whisper.cpp](https://github.com/ggml-org/whisper.cpp)'s
+bundled `whisper-server`, which can serve the OpenAI route directly:
+
+```bash
+brew install whisper-cpp
+whisper-server -m ggml-large-v3-turbo.bin \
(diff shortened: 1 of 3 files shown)


raleighbuckner:
Tested this branch on Windows + an NVIDIA RTX 5090 (Blackwell) using [speaches](https://github.com/speaches-ai/speaches) - the maintained successor to faster-whisper-server - as the OpenAI-compatible server, and it works end-to-end.

A data point for the review:

**Setup**
- speaches `latest-cuda` in Docker Desktop (WSL2 backend), model `Systran/faster-whisper-large-v3` loaded on the GPU
- `~/.config/watch/.env`:
  `WHISPER_BASE_URL=http://localhost:8000/v1`
  `WHISPER_MODEL=Systran/faster-whisper-large-v3`

**What worked**
- `/watch` on a captionless local video fell straight through to the local server - no API key, no `--whisper` flag needed.
- `WHISPER_MODEL` turned out to be essential here: speaches (unlike Groq/OpenAI's fixed model) rejects unknown model ids, and `model_override or …` handled it cleanly.
- `verbose_json` segments parsed unchanged, so timestamps survived.
- Nice touch that `_backend_label` prints `local` in the progress output instead of "groq"/"openai".

**One thing I noticed:** the progress line correctly says `sending to local Whisper`, but the final report header still prints `Source: whisper (groq)` - the report path uses the raw `used_backend` rather than `_backend_label`, so a fully-local run gets
labeled as Groq. Might be worth routing the report's source string through `_backend_label` too, for the same "don't imply it went to a third party" reason.

+1 either way - this really is the one hardcoded hostname standing between `watch` and fully-local transcription, and it's a clean change. Would love to see it land.

PR #76, opened 2026-07-15
Collapse the scrolled phase of YouTube's rolling captions (~49% fewer transcript tokens)

MindaugasJokubonis:
## The bug

`_dedupe` handles only the case where a cue contains **all** of its predecessor and extends it. YouTube's auto-caption cues cycle through three phases:

```
1. "little money. Um but life that's a very limited life."
2. "little money. Um but life that's a very limited life. Life can be much broader once you"
3. "Life can be much broader once you discover one simple fact. And that is"
```

Phase 2 is the handled case. **Phase 3 has scrolled** — it starts with the *tail* of phase 2, not the whole of it — so `startswith(prev + " ")` never fires and the repetition survives.

This looks like it's working: raw cues halve. But ~100% of the surviving segments still repeat their predecessor, so the transcript handed to the model is roughly **double** the actual speech — every duplicated word billed as input tokens and read twice by the model.

## Evidence

Measured on two real videos, English and Lithuanian:

| | segs | words | repeated |
|---|---:|---:|---:|
| **EN** raw cues | 64 | 724 | 100% |
| **EN** current `_dedupe` | 32 | 480 | 100% |
| **EN** with this patch | 17 | **245** | **0%** |
| **LT** current `_dedupe` | 522 | 5,419 | 85% |
| **LT** with this patch | 321 | **2,851** | **0%** |

**~49% fewer transcript tokens on English, ~47% on Lithuanian.** The Lithuanian word count independently matches an estimate of the unique content (2,798), so the collapse isn't eating real text.

## The fix

Phase 3 is the general case — longest suffix/prefix word overlap — and phase 1→2 falls out of it as "the overlap covers the whole previous cue", preserving existing behaviour there.

Two details:
- **Compared punctuation- and case-insensitively.** A scrolling cue re-emits the same word with different trailing punctuation (`blogai` → `blogai,`); an exact match misses it.
- **Overlap must be ≥ 2 words.** One shared word (`the`, `and`) is as likely to be coincidence as repetition, and dropping a legitimate word is a worse failure than leaving one duplicated.

## Tests

All 16 existing `tests/test_dedup.py` tests pass unchanged. Full suite shows no new failures (the 3 failures on my machine are pre-existing Windows issues on `main`, unrelated to this change).


MindaugasJokubonis:
Closing as a duplicate — I should have searched open PRs before opening this.

This is already covered by **#53** ("Auto-write TRANSCRIPT.md and fix rolling-caption overlap in dedupe"), which addresses the same `_dedupe` gap.

Cross-referencing the adjacent ones for anyone who finds this later:

- **#53** — rolling-caption overlap in `_dedupe` (this bug)
- **#30**, **#26**, **#75** — fetching non-English / original-language caption tracks rather than `en.*` only
- **#73** — capturing the yt-dlp exit code in `fetch_captions`

In case it's useful evidence for #53, measurements from the fix on real videos — English auto-captions went 480 → 245 words (49% fewer) and Lithuanian 5,419 → 2,851 (47%), with repeated segments dropping from ~100%/85% to 0% in both. The duplicated text is billed as input tokens and read twice by the model, so the impact isn't only cosmetic.

Sorry for the noise.

PR #77, opened 2026-07-15
Harden watch skill: security, robustness, tests (v0.3.0)

vcolombo:
## Context

A three-part review (security, functionality/robustness, contract/packaging) of the `watch` skill. The foundation was solid — list-form argv with `--` terminators, cert-verified TLS, no `shell=True`/`eval`/`pickle`, keys never logged — but residual issues clustered into injection/DoS hardening, hang/crash robustness, a latent auth bug, and an untested caption path. Scoped with the maintainer as **everything worth doing**, with input URLs treated as **untrusted**.

## Security (untrusted-input threat model)
- **SSRF guard** — `reject_internal_url` resolves the host and blocks loopback/private/link-local/reserved addresses (localhost, `169.254.169.254`, RFC1918) before yt-dlp runs.
- **Prompt-injection** — the skill's whole job is piping remote content into an agent, so a hostile caption/title carrying ` ``` ` + instructions is the real attack surface. Title/uploader/source are sanitized; the transcript is wrapped in a dynamic fence longer than any backtick run inside it and labeled *data, not instructions*.
- **Disk-fill DoS** — `--max-filesize` cap on downloads.
- Dropped the CWD `.env` key fallback; redacted Whisper error bodies to the structured `error.message` only; `chmod 0600` failures now warn instead of silently passing.

## Robustness
- Subprocess **timeouts** on every yt-dlp/ffmpeg/ffprobe call + live-stream rejection (no more infinite hang).
- **Audio-only / no-video-stream** sources degrade to transcript-only (was a crash that also discarded the transcript).
- `fetch_captions` surfaces yt-dlp's exit code (private/geo/removed) instead of masquerading as "no captions".
- `duration==0` no longer collapses to a single frame.
- token-burner **hard 500-frame ceiling**; long transcripts inlined as head/tail with the full text spilled to `transcript.txt`; native-language caption fallback; pre-download long-video heads-up.

## Cleanup / contract
- **Single shared `.env` parser** (`config.parse_env_file`/`env_value`) — fixes a latent auth bug where a trailing `# comment` broke only two of three divergent parsers, and deletes the duplication.
- OS-correct binary hints (point at `setup.py`, not `brew`).
- Added a "when to use" trigger clause to the description, aligned it across all four surfaces, agent-neutral wording, and fixed the hook's unexpandable `$CLAUDE_PLUGIN_ROOT` copy-paste command.
- Version `0.2.0` → `0.3.0`.

## Tests
- New `test_transcribe.py` (caption path had zero coverage).
- Locked the 83da59f inline-comment regression via the shared parser.
- Added `load_api_key` precedence + CWD-`.env`-ignored, SSRF guard, `auto_fps` math, `validate_range`, injection helpers, and Whisper `_error_summary` + 429-retry coverage.

## Verification
- All 7 scripts + 10 test files: `py_compile` + import clean, no dangling refs.
- **41-assertion stdlib self-check passed** covering every pure-logic path (shared parser, SSRF, download argv, VTT parse+dedupe, `auto_fps` duration-0, `validate_range`, injection helpers, key precedence + CWD-ignore, error redaction, 429 retry).
- ⚠️ **Not run here** (no ffmpeg/yt-dlp/pytest in the dev box): the full pytest suite and the ffmpeg/yt-dlp end-to-end paths. Run `python3 -m pytest -q` in an env with ffmpeg + yt-dlp before merge.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "name": "claude-video",
   "metadata": {
-    "description": "Give Claude a video input. /watch downloads, extracts frames, transcribes, and hands it all to Claude."
+    "description": "Give the agent a video input. /watch downloads, extracts frames, transcribes, and hands it all to the model."
   },
   "owner": {
     "name": "Bradley Bonanno"
@@ -9,7 +9,7 @@
   "plugins": [
     {
       "name": "watch",
-      "description": "Watch a video (URL or local path). Downloads with yt-dlp, extracts frames with ffmpeg, transcribes via captions or Whisper.",
+      "description": "Give the agent a video input (URL or local path). Downloads with yt-dlp, extracts auto-scaled frames with ffmpeg, and transcribes via native captions or the Whisper API.",
       "author": {
         "name": "Bradley Bonanno"
       },
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "watch",
-  "version": "0.2.0",
-  "description": "Let Claude watch a video. Downloads with yt-dlp, extracts auto-scaled frames with ffmpeg, pulls captions or falls back to Whisper, and hands frames + transcript to Claude so it can answer questions about the video.",
+  "version": "0.3.0",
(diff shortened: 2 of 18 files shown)


vcolombo:
Retargeting to the fork (vcolombo/claude-video) instead of upstream.

PR #95, opened 2026-07-27
Strip rolling-caption overlap: auto-generated transcripts are 2x larger than they need to be

jacobroberts0102-collab:
## Problem

YouTube's auto-generated caption tracks scroll, so each cue re-emits the tail of the cue before it. `parse_vtt` returns both copies, which roughly doubles the size of every transcript from a video without manual subtitles:

```text
[00:01] What would you say by the end of today's episode that the audience is going to
[00:03] episode that the audience is going to take away and what will they have
[00:04] take away and what will they have learned?
```

`_dedupe` already handles two rolling shapes, an exact repeat and a cue that grows by appending, but not this one, which is the shape `en-orig` tracks actually use. On a 55 minute video it is the difference between a 130KB and a 66KB transcript, and at `--detail transcript` the transcript is the entire report.

## Fix

`strip_rolling_overlap` drops the leading words of a cue that repeat the tail of the previous cue. Comparison is case and punctuation insensitive, since cues re-emit the same words with different capitalisation and commas. A 3-word minimum overlap (`MIN_OVERLAP_WORDS`) keeps genuine short repetition like "yeah, yeah" intact, and a cue left empty by the strip merges into its predecessor rather than disappearing.

It only runs on the caption path. `_dedupe` is reached solely through `parse_vtt`, so Whisper output is untouched.

## Verification

Measured on 8 real auto-caption files (7 different videos, 9 to 55 minutes, two channels):

| Check | Result |
|---|---|
| Transcript size | **50% smaller** on all 8, consistently |
| Nothing invented | output is an in-order subsequence of the raw cue stream |
| Nothing dropped | the `en-orig` and `en` tracks YouTube ships per video now agree on word count **exactly**, all 8 files |
| Idempotent | a second pass over deduped segments changes nothing |
| Manual captions | non-scrolling cues pass through verbatim |

That third row is the one I would lean on: two independently generated caption tracks for the same video converging on an identical word count after the strip is decent evidence the reconstruction is the real spoken text rather than an aggressive trim.

## Tests

`tests/test_transcribe.py` is new. The caption path had no test file, so it covers `parse_vtt` and friends generally, not just this change: all three rolling shapes, the overlap guard, case/punctuation insensitivity, idempotence, manual captions surviving verbatim, no spoken words dropped, inline tag stripping, plus `filter_range` and `format_transcript`. 14 tests, all passing.

Also documented in the Transcription section of SKILL.md.

## Note on the existing suite

5 tests fail on my machine (`test_setup.py::test_json_reports_watch_detail`, `test_keyless_completed_setup_proceeds_silently`, `test_watch.py::test_default_is_balanced`, `test_dedup_collapses_static_by_default`, `test_no_dedup_preserves_static_frames`). They fail identically on unmodified `main`, so they are not from this change. The cause is `WATCH_DETAIL=token-burner` in my own `~/.config/watch/.env` leaking into the tests, since `config.py` reads the real user config with no override for the test run. Happy to open that as a separate isolation issue if it is useful.


jacobroberts0102-collab:
Filed the pre-existing suite failures mentioned above as #96, with the root cause for each. None of them are from this change; they reproduce on unmodified `main`.

PR #97, opened 2026-07-27
Fix efficient-detail crash on keyframe-less ranges; silence Windows perms warning

sainbayare-net:
Two Windows-reproducible bugs found while installing the skill fresh on Windows 11 (Python 3.12, ffmpeg 8.1.1, yt-dlp 2026.07.04), plus the test-helper fix needed to see them.

## 1. `--detail efficient` aborts when a range contains no keyframes

```
$ python watch.py video.mp4 --detail efficient --start 12 --end 18
ffmpeg keyframe extraction failed: ... Could not open encoder before EOF
... Nothing was written into output file ... Conversion failed!
```

`extract_keyframes` raises on any non-zero ffmpeg exit. But when `-skip_frame nokey` selects nothing in the window, ffmpeg does **not** exit 0 with no output — it fails at mjpeg encoder init with *"No filtered frames for output stream"*. So the `len(candidates) < KEYFRAME_MIN` uniform fallback twelve lines below is unreachable in the zero-keyframe case, which is exactly the case it was written for.

This is easy to hit: encoders space keyframes seconds apart, so any short `--start/--end` window can land between two of them. `balanced` handles the identical window correctly (falls back to uniform), so the two modes disagree on the same input.

Fixed by treating an empty result as zero candidates and letting it flow into the existing fallback. A non-zero exit that *did* produce frames still raises, so genuine mid-run failures aren't swallowed, and a real decode failure still surfaces via the fallback's own ffmpeg call.

Before / after on the same command:

| | frames | exit |
|---|---|---|
| before | — | 1 (`Conversion failed!`) |
| after | 12 (uniform) | 0 |

The 12 frames match what `balanced` returns for that window.

## 2. `chmod 600` warning fires on every invocation on Windows

```
[watch] WARNING: C:\Users\...\.config\watch\.env is readable by other users.
Run: chmod 600 C:\Users\...\.config\watch\.env
```

Windows synthesizes `st_mode` as `0o666` regardless of the actual ACL, so `mode & 0o044` always matches. The warning cannot be dismissed (`chmod` isn't a Windows command, and `Path.chmod` there only toggles the read-only bit), and it prints on *every* `setup.py` call including `--check`.

That also breaks `test_keyless_completed_setup_proceeds_silently`, which asserts `--check` writes nothing to stderr.

Made the check POSIX-only. Access on Windows is governed by the user-profile ACL.

## 3. Test helper is blind to Windows paths

`_frame_lines()` in `tests/test_watch.py` matched only `/frames/frame_`, counting zero against the report's native backslash paths. Worth flagging that this cut both ways — `assert _frame_lines(out) == 1` would have passed vacuously if the real count had ever been zero.

## Verification

| | before | after |
|---|---|---|
| `pytest -q` (Windows) | 3 failed, 68 passed | **72 passed** |

Added `test_keyframe_fallback_when_range_has_no_keyframes`, which reproduces bug 1 on the repo's existing synthesized `static_clip` fixture — no network, no new fixtures. Confirmed non-vacuous: stashing the `frames.py` change makes it fail with the exact `SystemExit`.

**Caveat on coverage:** I only ran the suite on Windows 11. Bug 1 is platform-independent by nature — it's ffmpeg's behavior on an empty filter output, not Windows' — and the new test uses the existing fixture, so it should reproduce anywhere. But I haven't verified on macOS or Linux, and bugs 2 and 3 are Windows-only by construction. Worth a CI run before merging.

I left `CHANGELOG.md` alone since there's no `[Unreleased]` section — happy to add an entry wherever you'd prefer it.

Note: `hooks/scripts/check-setup.sh` has the same permissions blind spot as bug 2, but it's bash-only so I left it out of scope.


frinsen:
Thanks @sainbayare-net — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): merged (the frames half: a `--start/--end` window with no keyframe now falls back to uniform sampling instead of aborting). The Windows-permission half landed via #206.

PR #111, opened 2026-08-02
Fix three Windows-only defects: unclearable permission warning (x2) and POSIX-only frame assertions

iiiiins:
On Windows the test suite fails 3 of 71 on a clean checkout of `main`. Two root causes, three fixes.

## 1. The world-readable warning can never be cleared on Windows

`_check_file_permissions` reads `path.stat().st_mode`. On Windows CPython synthesizes that from the read-only attribute rather than reading the ACL, so every file reports group and other as readable and `mode & 0o044` is always true. The warning fires on **every** `/watch` invocation, and the remediation it prints does not help: `chmod` on Windows only toggles read-only, so `chmod 600` cannot clear the bits it is complaining about.

A warning that is a false positive 100% of the time on a platform is worse than no warning, because it trains people to scroll past the real one. This skips the check where `st_mode` carries no signal. Access on Windows is governed by NTFS ACLs on the user profile.

Verified against a real config file. Before:

```
[watch] WARNING: C:\Users\...\.config\watch\.env is readable by other users. Run: chmod 600 ...
```

After: silent, exit 0.

The function had no tests in either direction, so I added three: the warning still fires on a world-readable file under `posix`, stays silent at `0o600`, and is skipped under `nt`. Reverting only the `setup.py` change fails the `nt` case and passes the two `posix` cases, so the coverage does distinguish the fix from simply deleting the warning.

## 2. The same check exists a second time, in the SessionStart hook

`hooks/scripts/check-setup.sh` has its own copy using `stat -c '%a'`. Same root cause under Git Bash and Cygwin, and this is the copy users actually see, because it prints on every session start rather than only when `/watch` runs. On a Windows machine the banner reads:

```
/watch: WARNING — /c/Users/.../.config/watch/.env has permissions 644 (should be 600).
  Fix: chmod 600 /c/Users/.../.config/watch/.env
/watch: ready.
```

Gated on `$OSTYPE`. Verified both directions: silent under `msys`, and still warns on a 644 file with `OSTYPE` forced to `linux-gnu`.

## 3. `_frame_lines` counts zero on Windows

The helper matches the literal `"/frames/frame_"`, but the report prints native paths, so on Windows every line reads `...\frames\frame_0001.jpg` and the count is 0 regardless of what the extractor produced. `test_dedup_collapses_static_by_default` and `test_no_dedup_preserves_static_frames` fail for this reason alone.

Extraction itself is correct: the same static clip run directly yields 6 frames with the expected timestamps and `reason=uniform` tags. Normalizing separators before matching is the whole fix.

## Result

71 passed → 74 passed on Windows 11 / Python 3.13. No behavior change on macOS or Linux: both `posix` branches are untouched and now have coverage they lacked.

## One thing I did not change

`check-setup.sh:53` tells the user to run `python3 .../setup.py` when ffmpeg or yt-dlp is missing. On Windows `python3` is commonly a Microsoft Store stub that exits 49 without running anything, so that hint is dead on the machines most likely to need it. Left alone because picking the right interpreter is a broader call than this PR, but flagging it.


iiiiins:
Closing this — opened without the account owner's informed sign-off. The three Windows issues described are real and reproducible if you want to pick them up independently; nothing here needs to come from me.

PR #112, opened 2026-08-03
Add a local whisper.cpp backend for offline transcription

Diterex:
This adds a third transcription backend that runs fully on the local machine using whisper.cpp. No API key is needed and audio never leaves the computer.

## What it does

* New `WHISPER_CPP_BIN` and `WHISPER_CPP_MODEL` settings in the env file point at a `whisper-cli` executable and a ggml model file.
* `--whisper local` forces the local backend. With no flag the order stays the same as today, Groq first, then OpenAI, and now local as the last resort.
* When a cloud backend fails during a run and a local install is configured, the script retries locally so the user still gets a transcript.
* The setup preflight counts a configured local backend as ready, so keyless local users are not nagged.
* Docs updated in SKILL.md and the env template.

## Why

Some users cannot or do not want to send audio to a cloud API. Whisper is open source, so a local path gives the skill a zero key and zero cost mode. Existing cloud behavior is unchanged when a key is set.

## Testing

* Verified on Windows 10 with whisper.cpp v1.9.1 and the small English model. A real caption free video came back with a correct timestamped transcript tagged `whisper (local)`.
* The repo test suite gives the same results on this branch as on an unpatched checkout of main. 68 tests pass either way. The 3 that fail on both are environment related on my machine and are untouched by this change.


reddynitish:
Independently landed the same idea (local `whisper.cpp` fallback, same trigger condition — no captions, no cloud key) on a personal checkout and wanted to leave a real-world data point here rather than open a duplicate PR.

**Setup:** `brew install whisper-cpp` + `ggml-base.en.bin`, on an M4 Mac.

**Test:** an Instagram Reel with no native captions and no `GROQ_API_KEY`/`OPENAI_API_KEY` set. Ran `/watch` with zero flags.

**Result:** it auto-detected the missing cloud key, dropped to local `whisper-cli`, and transcribed a 66s clip in **1.8s** wall clock (Metal-accelerated). Output was clean — correct punctuation and capitalization out of the box with `base.en`, no priming prompt needed (see the note on #137 below).

One thing this PR's write-up doesn't mention that's worth folding in if you pick this up: #137 found that `whisper-cli`'s **timestamped** output can come out lowercase/unpunctuated on some models (they tested `large-v3` q5_0) unless primed with `--prompt "The following transcript uses standard punctuation and capitalization."`. I didn't reproduce that with `base.en` — properly punctuated by default — so it may be model-specific, but worth a defensive `--prompt` regardless since it's a no-cost flag per their testing.

Also worth noting: `whisper-cli --help` lists `mp3` as a directly supported input format (alongside wav/flac/ogg), so the existing `extract_audio()` → mp3 path can likely be reused as-is for the local backend without adding a separate wav-extraction step — that's what worked in my testing, contrary to the assumption in #137 that only wav is accepted.

Happy to share my diff if it's useful for comparison — mine's structured similarly (backend auto-falls-back to local when no cloud key is found, `--whisper local` to force it), just wanted to add confirmation this approach is solid and cheap to land.

PR #118, opened 2026-08-06
fix(windows): permission warning, console encoding, and two test-suite bugs

stickersfxlab:
Four Windows fixes found while running `/watch` v0.2.0 on Windows 11 (Git Bash, Python 3.14, ffmpeg 8.1, yt-dlp 2026.07.04).

On that platform the suite is **68 passed, 3 failed** before these commits and **71 passed** after. The temp-directory fix is not Windows-specific.

### 1. `.env` permission warning that cannot be satisfied

`_check_file_permissions` tests `st_mode & 0o044`, and `check-setup.sh` tests `stat -c %a != 600`. On Windows neither can ever pass:

- CPython synthesises `st_mode` from the read-only attribute alone; the group/other bits are always set regardless of the file's ACL.
- Under Git Bash on NTFS, `stat -c %a` returns `644` for any user file no matter what `chmod` did.

So both fired on every run and pointed at `chmod 600`, which cannot clear those bits. Restricting the file is an `icacls /inheritance:r /grant:r "%USERNAME%:F"` operation neither check can observe. Skipped on Windows; macOS and Linux behaviour is unchanged.

This is what breaks `test_keyless_completed_setup_proceeds_silently`, which asserts a silent `--check`.

### 2. Console encoding

Python still encodes stdout with the locale codepage on Windows (cp1252 here) while the agent harness reading it decodes UTF-8, so the `…` in the progress lines and the `—` in the report arrive mangled:

```
[watch] extracting scene-aware frames over full 6.0s (target 12, cap 100)?
_Work dir: `C:\...\watch-8e8lm9tp` ? delete when done._
```

`reconfigure(encoding="utf-8")` in all six scripts — SKILL.md invokes each one directly and each has a `__main__` block, so any of them can be the process writing to stdout. Full rationale is in `watch.py`; the other five carry a one-line reference.

### 3. `_frame_lines` counted zero frames on Windows

The helper matched `"/frames/frame_"`, but the report prints paths with the host separator, so every line reads `…\frames\frame_0001.jpg`. Both dedup assertions failed while dedup itself was working correctly — verified by hand on a 6 s static clip: `1 selected from 12 candidates … 11 near-duplicates dropped`, exactly what the test expects.

### 4. The suite leaks working directories

Every `watch.py` run without `--out-dir` calls `tempfile.mkdtemp()` and leaves the directory behind by design — the agent reads the frames from it and cleans up afterwards. The suite has no agent, so a full run strewed 12 `watch-*` directories through the real `TMPDIR` / `%TEMP%`. An autouse fixture points `TMPDIR`/`TEMP`/`TMP` at pytest's `tmp_path`, which already owns cleanup. Affects every platform.

### Verification

```
71 passed in 7.76s          # and 0 watch-* directories left behind
```

Also exercised end to end on Windows after the changes: a YouTube URL in `--detail transcript` mode (captions only, no video download, 6 transcript segments) and a local clip in the default `balanced` mode.


yejoopapa:
Confirming item 2 on a Korean-locale Windows install, where it is not cosmetic — it is fatal.

`watch.py` dies at the report print:

```
UnicodeEncodeError: 'cp949' codec can't encode character '—' in position 82: illegal multibyte sequence
```

The difference is codec coverage of the characters the report actually contains:

| char | cp1252 | cp949 |
|---|---|---|
| U+2014 EM DASH | `0x97` | **unmapped** |
| U+2015 HORIZONTAL BAR | unmapped | `0xa1aa` |
| U+2026 HORIZONTAL ELLIPSIS | `0x85` | `0xa1a6` |

CP949 inherits EUC-KR's punctuation block, which carries U+2015 HORIZONTAL BAR where cp1252 carries U+2014 EM DASH. So a Western Windows sees a mangled dash and continues; a Korean Windows raises and exits 1. The `…` in the progress lines survives on both — only the em-dash is fatal.

What makes this worse than the mangling suggests: the em-dashes are in the long-video warning and in the closing work-dir line, both of which print **after** the download, frame extraction and transcript parsing have already completed. The run does all the expensive work and then discards the entire report at the last step. Exit code is 1, so the caller sees a failure with no output at all.

Minimal repro on a cp949 console, no video needed:

```
$ python -c "print('a — b')"
UnicodeEncodeError: 'cp949' codec can't encode character '—' in position 2
$ echo $?
1
```

Verified end to end on a 14:53 Korean video (`--sub-langs` covering `ko`):

- before: exit 1, no report, 80 extracted frames orphaned on disk
- with `reconfigure(encoding="utf-8")` as in this PR: exit 0, report intact, no `PYTHONIOENCODING` needed

+1 on merging. Happy to add a regression test asserting the report survives a non-UTF-8 stdout if that would be useful.

PR #119, opened 2026-08-06
fix(windows): read .env in the encodings Windows actually writes

stickersfxlab:
All four `.env` readers in the package read the file as strict UTF-8. On Windows that is the one encoding the platform's own tools do *not* write by default, and the failure is worse than it looks: `UnicodeDecodeError` subclasses `ValueError`, **not** `OSError`, so it sails straight through the `except OSError` guard that was there to make config reading non-fatal.

Independent of #118 — this branch is cut from `main` and touches no line those commits touch.

### Four ways it breaks, measured on PowerShell 5.1.26100 / Windows 11

| How the file got written | Bytes | Result |
|---|---|---|
| `... \| Out-File .env` (PowerShell's default) | UTF-16LE + BOM | `UnicodeDecodeError` — crash |
| `Set-Content .env "# configuración"` | ANSI code page (cp1252) | `UnicodeDecodeError` — crash |
| UTF-16LE, no BOM | NUL-interleaved | decodes to junk, **every key silently vanishes** |
| Notepad, "UTF-8 with BOM" | `EF BB BF` prefix | first key becomes `﻿GROQ_API_KEY`, **silently ignored** |

Two crash and two lose the key without a word. This matters because `SKILL.md` instructs the agent to write `WATCH_DETAIL` and the API keys into this file, and an agent on Windows reaching for a shell redirect gets UTF-16 by default.

Concrete symptom for the silent pair: with a UTF-16LE+BOM `.env` that **did** contain a valid `GROQ_API_KEY`, `hooks/scripts/check-setup.sh` printed

```
/watch: ready for videos with native captions. Add GROQ_API_KEY (preferred) or OPENAI_API_KEY to ~/.config/watch/.env to unlock Whisper fallback.
```

— asking for a key that was already there, on every session, forever.

### The fix

`config.decode_env_bytes()` sniffs the BOM, falls back to UTF-16LE when an otherwise-ASCII file has interleaved NULs (a real `.env` never contains a NUL byte), and as a last resort decodes with `errors="replace"`, which cannot raise. ASCII keys and API keys survive a legacy code page intact; only the offending bytes become U+FFFD, with one line to stderr. Platform-neutral: nothing is gated on `sys.platform`.

**The site that actually matters is not `config.py`.** It is `setup.py:_read_env_key`, the preflight that runs on *every* `/watch` invocation — it crashed before `config.py` was ever consulted. So all three Python readers now share the one decoder rather than each doing its own `read_text`:

- `config.py:read_env_file` — the settings reader
- `setup.py:_read_env_key` — the preflight, and `_write_setup_complete`, which now also normalises the file back to UTF-8 on write so it stops being a hazard for everyone else
- `whisper.py:load_api_key._from_dotenv` — the API-key reader

The bash reader can't import Python, so `check-setup.sh` strips NULs and BOMs with `tr -d` before `awk`. Note the pipe forces dropping `awk`'s filename argument: with the file still named, `awk` re-reads the raw bytes and ignores the cleaned stream.

### Verification

```
main:            68 passed, 3 failed     # the 3 are the pre-existing Windows failures fixed by #118
this branch:     84 passed, 3 failed     # +16 new tests, all green
on top of #118:  87 passed
```

16 new tests, each verified to fail without its fix: a 6-way encoding parametrisation over `read_env_file`, a 4-way parametrisation over the `setup.py` preflight (`--check` exit 0 and `--json` reporting the right key/detail), a 4-way parametrisation over `whisper.load_api_key`, plus one asserting undecodable bytes never raise and one asserting an unreadable file still falls back to defaults.

The bash hook is not covered by pytest and was checked by hand: against the UTF-16LE+BOM `.env` above it now exits 0 in silence.


frinsen:
Thanks @stickersfxlab — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): adapted — `decode_env_bytes` (UTF-16 ± BOM, UTF-8 BOM, ANSI fallback with a warning) sits behind the single `.env` parser that #232 introduced, with four of your cases as tests.

PR #127, opened 2026-08-10
Fix YouTube returning nothing: retry with the android player client (SABR)

caleb436:
## Problem

YouTube forces SABR streaming on yt-dlp's `default` / `web` / `tv` / `ios` player clients. When it does, `/watch` returns **nothing at all** — the caption fetch and the media download both fail, so there is no transcript and no frames to fall back on.

On a video with manual English captions, both `fetch_captions()` and `download_url()` failed with:

```
ERROR: [youtube] <id>: The page needs to be reloaded.
```

and `download_url()` then raised `yt-dlp did not produce a video file`.

Because it takes out captions *and* media, a Whisper API key does not help either — there is no audio to transcribe.

## Measurement

The `android` player client bypasses SABR. I tested six clients against the same failing video, downloading a 3-second section of the same format string the script uses:

| player client | result |
|---|---|
| **`android`** | **OK** |
| `tv_embedded` | FAIL — "YouTube is no longer supported in this application or device" |
| `ios` | FAIL — requested format is not available |
| `mweb` | FAIL — requested format is not available |
| `web_safari` | FAIL — requested format is not available |
| default | FAIL — `HTTP Error 403: Forbidden` |

## Change

`fetch_captions()` and `download_url()` now try the **default client first**, and fall back to `android` only when the attempt produced nothing.

Default-first is deliberate: it selects better formats whenever SABR is not being forced, so this costs nothing on videos that already work. The two clients are never merged into a single run — with `youtube:player_client=default,android` the format picker re-selects the broken default formats and 403s again, so the retry has to be its own invocation.

Two supporting fixes, without which the fallback cannot fire:

- **`_pick_video()` rejects files under 1KB.** A 403 can leave a stub behind, and the previous check accepted any matching filename as a successful download — the retry would never trigger.
- **`_clear_media()` removes media and `.part` files before retrying.** Otherwise yt-dlp finds the previous attempt's output and reports it as already downloaded. It deliberately leaves `.info.json` and `.vtt` in place.

Both paths are gated on `is_youtube(url)`, so no other site pays for a second attempt.

## Testing

- `python3 -m py_compile` clean.
- Verified end to end on the failing video: 3 caption files, a 14MB `.mp4`, and frames extracted — where the same URL previously produced nothing.
- No behaviour change for non-YouTube URLs (the retry is unreachable for them) or for YouTube URLs where the default client already succeeds (the retry only runs when the first attempt produced nothing).

Happy to adjust the naming or split the `_pick_video` / `_clear_media` fixes into their own commit if you would prefer them separate.


frinsen:
Thanks @caleb436 — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): adapted — `android` is now in the media-fetch client fallbacks (`mweb`, `tv`, `web_embedded`, `android`), tried only when the default client is refused with 403/429/bot-check.

PR #131, opened 2026-08-11
Fix frames from a stale download when --out-dir is reused

NerdyHank:
## The problem

Reusing `--out-dir` across two different URLs returns the frames of the **first** video for the second run, while every text field in the report correctly describes the second one.

```bash
python3 skills/watch/scripts/watch.py "https://www.youtube.com/watch?v=AAAAAAAAAAA" --out-dir /tmp/w
python3 skills/watch/scripts/watch.py "https://www.youtube.com/watch?v=BBBBBBBBBBB" --out-dir /tmp/w   # frames are still A
```

`download/video.mp4` from the first run is still there, so yt-dlp reports `has already been downloaded` and skips fetching — but it *does* rewrite `video.info.json` and the subtitles. Title, uploader and transcript therefore come from B while ffmpeg extracts frames from A's media file.

This is nastier than a normal cache bug because the run looks entirely successful. Nothing in the report is wrong except the images, and the images are the part a model consumes without a human ever looking at them. We hit it through a downstream skill that reuses one working directory: it produced a finished, illustrated document in which every screenshot came from an unrelated video.

`Duration` is the one field that gives it away — it is probed from the media file, not taken from the provider metadata — but you only notice if you happen to know how long the video should be.

## The fix

Three independent guards, so neither the cache nor the caller can set the trap:

**1. One cache directory per source.** Downloads land in `download/<source_key>/` instead of a shared `download/`. Re-running the same URL still hits the same directory, so `--out-dir` keeps working as a cache; a different URL structurally cannot inherit another's media. Both `download_url()` and `fetch_captions()` resolve the same directory, so the captions pass and the video pass never disagree on location.

`source_key()` canonicalises the URL first:

| URL | key |
|---|---|
| `watch?v=X` | `youtube:X` |
| `watch?v=X&t=90s` | `youtube:X` |
| `youtu.be/X?si=…` | `youtube:X` |
| `shorts/X` | `youtube:X` |

Only positional/referrer params (`t`, `si`, `list`, …) are dropped. On unknown hosts every meaningful query parameter is kept, so two distinct videos can never collapse into one entry — worst case is a redundant re-download, never a wrong one.

**2. Cross-check after every fetch.** `_duration_mismatch()` measures the media with ffprobe and compares it against the duration in `video.info.json`, tolerance `max(8s, 3%)`. On disagreement the directory is discarded and fetched once more from scratch; if it still disagrees, a loud warning instead of quietly wrong frames.

**3. Transcript-vs-media check.** Captions cannot run past the end of their own video. If the last cue is more than 30s beyond the media duration, the report gets a `STOP — mismatched sources` banner, placed inline where the frame list is read.

Also adds `--fresh` to discard a cached download on demand.

## Tests

Nine tests in `tests/test_download.py`, no network — canonicalisation across URL spellings, distinct keys for distinct videos, no false merge on unknown hosts, two URLs not sharing a directory, captions and download agreeing on one directory, `--fresh`, and the duration detector in both directions including a missing `info.json`.

```
80 passed in 2.74s
```

## Verification

End-to-end against this branch — two different videos through one `--out-dir`:

```
Run 1  DIY Anemometer …          15:15 (915.3s)
Run 2  Master the NEW Gemini …   20:45 (1245.4s)   # same --out-dir

download/267bf458946a/video.mp4   915.334000
download/90958a4fdb49/video.mp4  1245.434000
```

Before the change, run 2 reported the second title with the first video's frames. Now both the metadata and a frame read out of run 2 show the correct video.

Also checked by hand: cache hit via a different URL spelling of the same source (no re-download, no third directory), a foreign file planted inside the correct cache directory (detected, discarded, re-fetched), and the local-file path.

No new dependencies — ffprobe was already required.

Rebased onto current `main` (0.2.0 package layout); the guards cover the `fetch_captions()` path introduced there as well.


NerdyHank:
Withdrawing this — not pursuing it upstream. Leaving the analysis here in case it's useful to anyone hitting the same thing: reusing --out-dir across two different URLs lets the second run inherit the first video's media, because yt-dlp skips the download but still rewrites info.json and the subtitles. The report then reads correct while every frame is from the other video.

PR #136, opened 2026-08-12
fix(setup): preflight must read cwd/.env like whisper.py does

gqbeerman:
## The bug

`whisper.load_api_key` searches, in order:

1. `$GROQ_API_KEY` / `$OPENAI_API_KEY`
2. `~/.config/watch/.env`
3. `cwd/.env`

`setup._read_env_key` stops at the first two.

So a key kept in a project-local `.env` — which is where a lot of us keep per-project credentials — transcribes perfectly, while `setup.py --check` prints `setup incomplete (no Whisper API key ...)` on **every single run** and `--json` reports `status: "needs_key"`. The remedy it suggests, running setup to add a key, can't help: the key is already there and already working.

I hit this vendoring `watch` into a project whose `.env` sits at the repo root. Transcription worked on the first try; preflight insisted it wasn't configured.

## The fix

`_read_env_key` walks the same paths in the same order as `whisper.py`. Two side effects of the old early `return` also go away:

- an empty `GROQ_API_KEY=` in `~/.config/watch/.env` no longer masks a real key in `cwd/.env`
- an unreadable file is skipped rather than aborting the whole search

Precedence is unchanged otherwise: env var wins, then config file, then cwd — and backend preference (Groq before OpenAI) still dominates file location, since each key name is searched across all paths before the next name is tried.

## Tests

Three added, **each one fails before this change**:

- `test_key_in_project_dotenv_is_ready` — the reported bug
- `test_empty_config_value_falls_through_to_project_dotenv` — the masking case
- `test_backend_preference_beats_file_location` — pins the existing semantics so this doesn't regress

The `_run` helper gains a `cwd` parameter so the working directory can be controlled.

```
74 passed   (71 on main)
```

Thanks for the skill — the frame extraction and the dedup pass in particular are doing real work for me.


frinsen:
Thanks @gqbeerman — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): adapted — `setup.py`'s preflight now uses the same config → `./.env` search order as `whisper.py`, via the shared `read_env_value`.

PR #147, opened 2026-08-16
Use a sidecar .vtt as the transcript for local video files

JMAL1988:
## Problem

`download.resolve_local()` returns `subtitle_path: None` unconditionally, so a local video file can only get a transcript via the Whisper API path in `whisper.py` (Groq or OpenAI). With no key in `~/.config/watch/.env`, `/watch ./clip.mp4` is frames-only — no spoken content, ever.

That is the one source type that can never have platform captions, and it is also the only one with no free transcript route. On a machine that can transcribe locally (or where a `.vtt` already exists next to the file — an editor export, a previous run, a subtitle download) the transcript is right there and the skill ignores it.

## Change

`resolve_local()` now looks for a VTT sitting next to the video:

1. exact sibling — `clip.mp4` -> `clip.vtt`
2. language-tagged sibling — `clip.en.vtt`, `clip.nl.vtt` (the shape yt-dlp writes)

Nothing else changes. URL handling is untouched, platform captions still take priority where they exist, and the Whisper API fallback still runs when no sidecar is found. `SKILL.md` documents it as the second of three transcript routes.

## Why this shape

It keeps the skill pure-stdlib and adds no dependency or config. Any local transcription tool — `whisper`, `faster-whisper`, `whisper.cpp`, all of which write VTT — becomes a usable backend by writing its output next to the video, without the skill having to pick or vendor one.

## Tests

Four cases added to `tests/test_download.py`: no sidecar, exact sidecar, language-tagged sidecar, and an unrelated `.vtt` in the same directory (must not match).

```
$ python3 -m pytest tests -q
75 passed in 4.95s
```

## Where this came from

Ran into it using `/watch` on local rushes and screen recordings on a workstation with a local GPU: captions never apply, and paying an API to transcribe a file already sitting on the same box that could do it for free was the wrong trade. The local transcription itself stays out of the skill — this PR only opens the door for it.


frinsen:
Thanks @JMAL1988 — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): merged as authored — `clip.vtt` / `clip.<lang>.vtt` next to a local file is now the transcript.

PR #154, opened 2026-08-17
docs(skills): make watch description state when to invoke

thomaswillner:
The `watch` skill description currently spends most of its length on implementation mechanism — yt-dlp, ffmpeg, the Whisper fallback — rather than on when an agent should invoke it.

Per the [Agent Skills specification](https://agentskills.io/specification), the description is the only thing an agent sees before deciding whether to load a skill. Naming the tooling does not help that decision; naming the user intents does. The mechanism is already documented in the skill body, which loads on invoke.

**Before**
```yaml
description: Watch a video (URL or local path). Downloads with yt-dlp, extracts auto-scaled frames with ffmpeg, pulls the transcript from captions (or Whisper API fallback), and hands the result to Claude so it can answer questions about what's in the video.
```

**After**
```yaml
description: Watch a video from a URL or local path and answer questions about its content. Use when the user shares a video and asks what is in it, wants it summarised, or asks about something shown or said on screen.
```

Shorter, and it now lists the distinct intents that should trigger it. One line changed; no behavioural change.


frinsen:
Thanks @thomaswillner — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): merged as authored — the description now says *when* to invoke, which is what hosts match on.

Issue #163, opened 2026-08-21
Frame extraction fails on ffmpeg 7+: `-vsync` was removed in favour of `-fps_mode`


dmitrykarabutov:
`skills/watch/scripts/frames.py` passes `-vsync vfr` to ffmpeg on lines 256 and 615. That option was deprecated and then removed; current ffmpeg rejects it outright, so scene extraction fails and `/watch` returns zero frames.

Observed on macOS 15 (Apple Silicon), ffmpeg 9.0.1 installed via Homebrew, watch 0.2.0.

Output:

```
[watch] extracting scene-aware frames over full 13.0s (target 13, cap 100)…
ffmpeg scene extraction failed: Unrecognized option 'vsync'.
Error splitting the argument list: Option not found
```

The run still exits 0 and reports success, so it looks like the video simply had no extractable frames rather than like a failure — that part cost more time to diagnose than the flag itself.

Replacing `-vsync vfr` with `-fps_mode vfr` in both places fixes it; I've been running patched locally with no issues. `-fps_mode` has been supported since ffmpeg 5.1, so it may be safe unconditionally, though you may prefer to detect the version if you support older installs.

Happy to open a PR if useful.

Issue #161, opened 2026-08-20
Frame extraction fails on ffmpeg 9.0: -vsync was removed (use -fps_mode)


badwookie208:
Frame extraction fails completely on **ffmpeg 9.0**, which removed the deprecated `-vsync` option. The result is transcript-only output with no frames — which for a skill whose purpose is giving Claude a *visual* input is a total failure rather than a degradation.

Notably this hits **new Windows users following the plugin's own setup instructions**: `scripts/setup.py` prints `winget install Gyan.FFmpeg`, and that package now ships **9.0**. So a first run on a clean Windows box installs an ffmpeg the extractor cannot drive.

## What happens

```
[watch] extracting scene-aware frames over full 421.4s (target 60, cap 60)…
ffmpeg scene extraction failed: Unrecognized option 'vsync'.
Error splitting the argument list: Option not found
```

Download and caption retrieval both succeed; only frame extraction dies.

## Cause

`scripts/frames.py` passes `-vsync vfr` in two places:

- `frames.py:256`
- `frames.py:615`

`-vsync` was long deprecated in favour of `-fps_mode` and is **removed** in ffmpeg 9.0, so the whole command is rejected at argument-parsing time.

## Reproduction (self-contained, no video needed)

```bash
# a source with real scene cuts
ffmpeg -f lavfi -i "testsrc=duration=2:size=160x90:rate=10" \
       -f lavfi -i "smptebars=duration=2:size=160x90:rate=10" \
       -f lavfi -i "color=c=navy:duration=2:size=160x90:rate=10" \
       -filter_complex "[0:v][1:v][2:v]concat=n=3:v=1[v]" -map "[v]" -y scenes.mp4

# what frames.py sends today
ffmpeg -i scenes.mp4 -vf "select='gt(scene,0.1)',scale=128:-2" -vsync vfr a_%02d.jpg -y
#   exit 8 — "Unrecognized option 'vsync'"  →  0 frames

# the replacement
ffmpeg -i scenes.mp4 -vf "select='gt(scene,0.1)',scale=128:-2" -fps_mode vfr b_%02d.jpg -y
#   exit 0  →  2 frames
```

## Fix

Swapping `-vsync vfr` → `-fps_mode vfr` at both call sites resolves it. I ran the full pipeline on a 7-minute YouTube video that way and got frames as expected.

[...]

Issue #156, opened 2026-08-18
YouTube media 403 on Windows despite deno + curl_cffi present: only player_client=mweb works, but mweb alone kills subtitles


realsalestalk:
## Summary

On Windows with a fully-provisioned environment (deno present, curl_cffi impersonation targets available), every YouTube **media** download returns `HTTP Error 403: Forbidden` while captions download fine in the same run. The only `player_client` that returns usable media URLs is `mweb`.

This is **not** #93 (brew yt-dlp with zero impersonation targets) or #67 (missing deno/curl_cffi) — both of those root causes are absent here, and `--impersonate` does not help. Filing separately since the remedy is different.

## Environment

- Windows 11, Python 3.14.5
- yt-dlp `nightly@2026.08.17.073947` (stable `2026.07.04` behaves identically)
- ffmpeg `N-125875-g5d4d3bdc61-20260731`
- `yt-dlp --list-impersonate-targets` → Chrome-133/136, Safari-17.2/18.0/18.4/26.0, etc. (curl_cffi bundled in the standalone exe)
- `deno` installed and on PATH

## What fails

`yt-dlp -F` lists formats fine, so the failure is specific to fetching media, not to extraction:

```
ERROR: unable to download video data: HTTP Error 403: Forbidden
```

Reproduces on any video, including a 19-second unrelated one, so it is not video-specific or rate limiting.

Tested per client on `-f "worst[height<=360]/worst"`:

| player_client | media download |
|---|---|
| `default` | 403 |
| `web` | 403 |
| `tv` | 403 |
| `web_safari` | 403 |
| `ios` | 403 |
| **`mweb`** | **OK** |

`--impersonate chrome` and `--impersonate safari` also 403 with the default client.

## The part that makes the obvious fix wrong

Setting `player_client=mweb` globally fixes media but **silently breaks subtitles** — mweb requires a PO token for them:

```
WARNING: [youtube] There are missing subtitles languages because a PO token was
not provided. Subtitles for these languages are missing: en, ...
```

So `/watch` then returns 100 frames and `Transcript: none available`, which reads like a captions bug rather than a side effect.

Listing both clients does not resolve it either — it just flips which half breaks:

| extractor-args | subtitles | media |
|---|---|---|
| `player_client=mweb` | missing | OK |
| `player_client=default,mweb` | OK | 403 |
| `player_client=default` | OK | 403 |

## Fix that worked

[...]

Issue #150, opened 2026-08-16
UnicodeEncodeError on Windows: report printing crashes on cp1252 in focused mode


mkgaskin-ops:
**Environment**
- Windows 11, PowerShell, console codepage cp1252
- Python 3.14.3, yt-dlp 2026.08.16 nightly, ffmpeg N-124716
- Claude Code v2.1.222, skill installed manually to ~/.claude/skills/watch

**What happens**

Frame extraction, dedup, and transcript all complete successfully. The run then dies while printing its own report:

```
UnicodeEncodeError, watch.py:280
```

**Cause**

`watch.py:280` emits U+2192 RIGHTWARDS ARROW in the Focus range line:

```python
f"- **Focus range:** {format_time(effective_start)} \u2192 {format_time(effective_end)} "
```

Python's stdout defaults to cp1252 on Windows, which cannot encode it. Because that line sits inside `if focused:`, the crash only reproduces when `--start` or `--end` is passed; full-video runs skip it.

Note there is also an em dash at `watch.py:266` on the "no audio stream found" stderr path, so escaping the arrow alone would leave a second landmine.

**Repro**

```
/watch <any URL> --detail efficient --start 0 --end 10 <question>
```

**Suggested fix**

Force UTF-8 at the top of `main()` rather than escaping individual characters:

```python
if hasattr(sys.stdout, "reconfigure"):
    sys.stdout.reconfigure(encoding="utf-8", errors="replace")
    sys.stderr.reconfigure(encoding="utf-8", errors="replace")
```

**Workaround**

`PYTHONIOENCODING=utf-8`, or `setx PYTHONUTF8 1` for the user account.

PR #167, opened 2026-08-25
Fix two Windows blockers: ffmpeg 8+ -vsync removal and cp1252 UnicodeEncodeError

MadPonyInteractive:
Two independent bugs that together make `/watch` unusable on a current Windows install. Both are already reported several times over; this is the fix.

## 1. `-vsync` was removed from ffmpeg

`frames.py` passes `-vsync vfr` on both extraction paths. ffmpeg removed the option, so extraction aborts:

```
Unrecognized option 'vsync'.
```

The skill surfaces this as `ffmpeg scene extraction failed` **after** the video has finished downloading, so a run burns the full download and returns nothing.

`-fps_mode` is the direct replacement and has been supported since ffmpeg 5.0, which is older than the ffmpeg this skill's own installer pulls on every platform. There is no version window that accepts `-vsync` but not `-fps_mode`, so no fallback is needed.

Reported in #163, #161, #149, #143, #141, #134, #126, #122, #117.

## 2. `UnicodeEncodeError` on cp1252 consoles

`watch.py`, `frames.py`, `setup.py` and `whisper.py` all print em dashes, ellipses and arrows. A Windows console is cp1252 by default and cannot encode any of them:

```
UnicodeEncodeError: 'charmap' codec can't encode character '→'
```

The most reliable trigger is the focus-range line, printed with a real `→`, so **every** `--start`/`--end` run crashes. Like the first bug it fires after frames are extracted, which makes it read as a frame failure rather than a printing one.

Reconfiguring `sys.stdout` / `sys.stderr` in the two entry points covers the whole skill — `watch.py`'s reconfigure applies to `frames.py`, `whisper.py`, `transcribe.py` and `download.py` as well, since they print to those same streams. `setup.py` gets its own because it is invoked directly.

`errors="replace"` so an odd character degrades to a placeholder instead of taking the run down, and the call is guarded because `sys.stdout` is not always a `TextIOWrapper` when the skill runs under a harness that has swapped the stream.

Reported in #150, #109, and the second half of #134.

## Not covered here

Deliberately left alone so this stays reviewable:

- #108 is a `UnicodeDecodeError` **decoding subprocess output**, which is the opposite direction to this and needs its own fix.
- #128 (App Control blocking `ffprobe.exe`), #156 (YouTube 403 / player_client) and #107 (NTFS ACL permission check) are unrelated.

## Verification

Windows 11, Python 3.14, ffmpeg 9.0-full_build, yt-dlp 2026.08.19.

Before: `/watch` on a YouTube URL downloads the video, then dies on `Unrecognized option 'vsync'`. Patching only that, a `--start 0 --end 10` run then dies on the arrow while printing the report.

After: both scene-aware and keyframe extraction produce frames, the focus-range line prints, and `setup.py --json` completes.

```
- **Focus range:** 00:00 → 00:10 (10.0s)
- **Detail:** balanced
- **Frames:** 7 selected from 3 candidates (uniform with uniform fallback,
  13 near-duplicates dropped, focused range, budget 20, cap 100)
```

21 insertions, 2 deletions, across 3 files. No behaviour change on macOS or Linux: `-fps_mode` is equivalent to `-vsync` there, and reconfiguring an already-UTF-8 stream to UTF-8 is a no-op.


MadPonyInteractive:
Closing this - the same fixes are already covered by several existing reports, so an extra PR isn't adding anything. Leaving it to the maintainers. For anyone landing here from a search: the two fixes are `-vsync` -> `-fps_mode` in skills/watch/scripts/frames.py, and reconfiguring sys.stdout/sys.stderr to UTF-8 in watch.py and setup.py for cp1252 Windows consoles.

PR #175, opened 2026-08-28
Compare dedup thumbnails in RGB, not grayscale

Daily-AC:
## The bug

`_thumb_frames` decodes dedup thumbnails with ffmpeg's `format=gray`, so two shots whose hues differ but whose **luma matches** compare as identical — and the second one is deleted as a near-duplicate.

A hard cut from a red card to a green card is the clearest case: `#FF0000` has luma 76.2, `#008200` has luma 76.3. In grayscale the pair measures a **1.0** mean delta against the **2.0** threshold — inside it, so the incoming shot is dropped and the model never learns it existed.

This isn't exotic footage. Motion-graphics pieces, title cards and brand reels cut between saturated colour fields constantly, and that's exactly the material where losing a shot changes the answer.

## Repro

12s clip: red (3s) / green (3s) / a held static frame (6s).

```bash
ffmpeg -f lavfi -i "color=c=0xFF0000:s=640x360:r=25:d=3" \
       -f lavfi -i "color=c=0x008200:s=640x360:r=25:d=3" \
       -f lavfi -i "testsrc2=s=640x360:r=25:d=6" \
       -filter_complex "[2:v]select='eq(n,0)',loop=loop=149:size=1:start=0,setpts=N/25/TB[held];[0:v][1:v][held]concat=n=3:v=1[out]" \
       -map "[out]" -pix_fmt yuv420p -t 12 shots.mp4

python3 skills/watch/scripts/watch.py shots.mp4 --detail balanced --no-whisper
```

| | frames kept | dominant colours |
|---|---|---|
| before | 2 | `#FE0000`, `#78847C` |
| after | 3 | `#FE0000`, `#008300`, `#78847C` |

Before this patch the entire 3-second green shot is gone.

## The fix

Decode to `rgb24` and size each thumbnail 3 bytes per pixel.

`_frame_delta` is deliberately **unchanged** — it already averages absolute byte differences, which becomes a per-channel mean on the same 0-255 scale. So `DEDUP_THRESHOLD` stays at 2.0 and nothing needs retuning. Measured on both edge cases:

| case | before | after | threshold |
|---|---|---|---|
| equal-luma cut (must be kept) | 1.000 ❌ | **128.333** ✅ | 2.0 |
| near-duplicate (must collapse) | 0.008 ✅ | **0.082** ✅ | 2.0 |

The near-duplicate margin stays wide, and in the repro above the static section still collapses 9 near-duplicates — the dedup savings this feature exists for are intact. Cost is 3x the thumbnail bytes, which at `DEDUP_THUMB = 16` is 768 bytes per frame instead of 256.

Docstrings and the constant comment are updated to match.


frinsen:
Thanks @Daily-AC — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): merged as authored (one test updated for the ×3 thumbnail size).

PR #176, opened 2026-08-28
Correct the image-token numbers in the /watch skill

Daily-AC:
The **Token efficiency** section currently says:

> 80 frames at 512px wide is roughly 50-80k image tokens depending on aspect ratio.

Claude prices images in 28x28-pixel patches — `ceil(width / 28) * ceil(height / 28)` visual tokens ([vision docs](https://platform.claude.com/docs/en/build-with-claude/vision#resolution-and-token-cost)). At the default `--resolution 512`, a 16:9 frame is 512x288:

```
ceil(512/28) * ceil(288/28) = 19 * 11 = 209 tokens
209 * 80 = 16,720
```

So the real figure is **~17k**, not 50-80k — the doc overstates the skill's main cost by 3-5x. The formula reproduces all three worked examples in the vision docs exactly (200x200 = 64, 1000x1000 = 1296, 1092x1092 = 1521), so it's not an approximation.

This matters because it's the number a model reads before deciding whether it can afford a full-fidelity pass. Overstating it pushes toward `--detail efficient` or a lower `--max-frames` when `balanced` was affordable all along.

Changes:

- Replace the fixed range with the formula plus the derived numbers, so the estimate stays right when someone changes `--resolution` or `--max-frames`.
- The "1024 roughly quadruples" line was correct (`37 * 21 = 777`, 3.7x) — kept, with the arithmetic shown.
- Add a note that tiling frames into contact sheets saves nothing. Cost follows pixel area, so nine frames in one 3x3 sheet cost the same as nine separate images (they're within 1% — the sheet is marginally *worse* once you add label strips). It's a natural-looking optimisation that does nothing, and the inflated number made it look necessary.


frinsen:
Thanks @Daily-AC — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): merged as authored — the 28-px patch arithmetic replaced the old "50–80k" estimate in SKILL.md.

PR #179, opened 2026-08-30
watch: fall back to android client (format 18) on download 403

pornthepp:
## Problem
YouTube's SABR rollout now requires a GVS PO token for most formats. Without one, the primary `yt-dlp` download in `skills/watch/scripts/download.py` fails with `HTTP Error 403: Forbidden` on many videos, even though caption/metadata fetching still succeeds.

## Fix
When the primary download produces no video file, retry once with `--extractor-args youtube:player_client=android -f 18`. The android client is currently exempt from the PO-token requirement but only exposes a single muxed format (itag 18, 360p mp4 + aac). This lets `/watch` degrade gracefully to lower-res frames instead of failing outright.

## Testing
Reproduced the 403 on a real video (`https://www.youtube.com/watch?v=mFgRGSOGNPM`) with yt-dlp 2025.10.14, confirmed the primary format string fails, and confirmed the android/format-18 fallback downloads successfully end-to-end through `watch.py` (captions + 82 scene-aware frames extracted).

Also tried `android_vr` (lists 1080p+ without a PO-token warning) and a self-hosted bgutil PO-token provider — both still 403 on the actual CDN fetch, so android/format-18 is the most reliable fallback currently available without cookies.


frinsen:
Thanks @pornthepp — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) (main, after [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0)): adapted — the `android` client fallback is in; the forced format-18 (360p) downgrade was left out so a successful retry keeps the normal quality ceiling.

PR #181, opened 2026-08-30
Fix frame extraction failing on ffmpeg 9 (-vsync removed)

MaCeeeee:
## The bug

ffmpeg 9.0 removed the deprecated `-vsync` option. Every frame-extraction engine in `frames.py` passes it, so all of them abort before writing a frame:

```
[watch] extracting keyframes over 01:00-01:10 (10.0s) (target 20, cap 50)…
ffmpeg keyframe extraction failed: Unrecognized option 'vsync'.
```

This is easy to miss because the transcript path is unaffected. On a video without captions the run reports "No transcript available" and no frames, which reads like a video that simply has nothing to offer — not a broken install. `--detail transcript` looks entirely healthy.

The repo's own suite catches it: **21 of 71 tests fail on ffmpeg 9** (`test_frames`, `test_dedup`, `test_timestamps`, `test_watch`).

Homebrew moved `ffmpeg` to 9.x, so new macOS installs hit this on their first `/watch` with frames.

## Why not just swap in `-fps_mode`

That trades one break for another. `-fps_mode` only exists from ffmpeg 5.1 on, and Ubuntu 22.04 LTS — in standard support until 2027 — ships 4.4. Neither spelling covers every ffmpeg the skill installs onto, and `setup.py` installs whatever the platform's package manager offers rather than pinning a version.

## The fix

`_vfr_args()`, an `lru_cache`d capability probe that asks ffmpeg what it supports and returns `-fps_mode vfr` or `-vsync vfr` to match. Used at both call sites (scene candidates, keyframes).

Two details worth flagging for review:

- **Probe, don't parse `-version`.** Git builds report `N-109848-g4b5b8b1`, which carries no usable version number.
- **`-h long`, not `-h full`.** Both list the option; `-h long` is ~14 KB against ~1 MB. One subprocess per process, cached.

Stays pure-stdlib (`functools.lru_cache`), consistent with the rest of the package.

## Testing

| | ffmpeg 9.0.1 (macOS arm64) |
|---|---|
| before | **21 failed**, 50 passed |
| after | **72 passed** |

The added test isn't a unit test of the probe — it feeds the resolved flag to ffmpeg for real, so if a future release removes this spelling too, the suite fails instead of the user's next render.

Both fallback branches were exercised by stubbing the probe:

```
real (ffmpeg 9):        ('-fps_mode', 'vfr')
simulated (ffmpeg 4.4): ('-vsync', 'vfr')
no ffmpeg at all:       ('-fps_mode', 'vfr')
```

End-to-end after the fix, frames land and are readable as images:

```
- **Frames:** 4 selected from 4 candidates (keyframe, focused range, budget 20, cap 50)
- frame_0001.jpg (t=01:00, reason=keyframe)  …
```

I left `CHANGELOG.md` alone — 83da59f suggests entries are written at release time rather than per fix. Happy to add one if you'd rather have it in the PR.


okami77-dev:
Verified this branch on macOS with Homebrew, which now ships ffmpeg 9, so `/watch` extracts no frames there on `main`.

Setup: `watch.py` from `main` (83da59f) and from this branch (84e05c8), run on a synthetic 24 s H.264 clip with 12 hard cuts every 2 s, in `balanced` and `efficient` (`--max-frames 40 --no-whisper`).

| ffmpeg | `main` | this PR |
|---|---|---|
| 9.0.1 (Homebrew `ffmpeg`) | exit 1 in both modes: `Unrecognized option 'vsync'` | probe picks `-fps_mode vfr`, 12/12 frames in both modes |
| 5.1.10 (Homebrew `ffmpeg@5`) | works via `-vsync` | probe picks `-fps_mode vfr`, same 12 frames |
| simulated build older than 5.1 ¹ | works via `-vsync` | probe picks `-vsync vfr`, same 12 frames |

In every successful run the frame timestamps fall exactly on the cuts (0, 2, …, 22 s), so the `showinfo` timestamps still line up with the files written under `-fps_mode vfr`. `ffmpeg -h long` lists `-fps_mode` on both 9.0.1 and 5.1.10, so the lighter probe is enough.

¹ There is no real ffmpeg older than 5.1 on this machine. I used a wrapper around 5.1.10 that removes `fps_mode` from `-h` output and rejects `-fps_mode` with ffmpeg's own `Unrecognized option` error.

#219 behaved identically in the same matrix, so either PR fixes this. I did not run the pytest suite.

PR #182, opened 2026-08-31
Fix rolling-caption duplication and dropped cues in VTT parsing

eltokhy:
`/watch` is great, thanks for building it. While using it on an 18:30 conference talk I noticed the transcript was about twice as long as the talk. Digging in turned up two bugs in `transcribe.py`, both specific to YouTube auto-subs.

## 1. Rolling duplication was never collapsed

`_dedupe` only caught a cue that repeated the previous one **in full**:

```python
if out and seg["text"].startswith(out[-1]["text"] + " "):
```

But a rolling cue starts with the previous cue's **tail**, not the whole of it:

```
cue N:   ">> Hi everyone. My name is Luke and my goal"
cue N+1: ">> Hi everyone. My name is Luke and my goal" + "is that 20 minutes from now you'll be"
cue N+2: "is that 20 minutes from now you'll be"       + "able to assemble agent teams that can"
```

`startswith` never matched, so every carry-over line was emitted twice. On [this 18:30 talk](https://www.youtube.com/watch?v=ow1we5PzK-o) that produced **5,975 words for 3,035 words of speech**, and left **465 of 523 consecutive segment pairs** sharing a 3-or-more word run. The transcript is surfaced in every captioned mode, so it is a real token cost on long videos.

## 2. Cues with a padding line parsed as empty

The cue body ended at the first line that stripped to nothing:

```python
while i < len(lines) and lines[i].strip():
```

YouTube pads cue bodies with a single-space line before the content. At the byte level the cue separator is `''` while the padding is `' '`, so the loop exited immediately and those cues parsed as empty:

```
00:00:15.040 --> 00:00:18.230 align:start position:0%
' '                                     <- loop stops here
'&gt;&gt; Hi<00:00:15.160><c> everyone.</c>…'   <- never read
```

Mid-video this mostly hides, because the dropped text reappears as the next cue's carry-over. At the **edges of the track there is no neighbour to carry it**, so it is lost outright: on this talk the transcript began at `[00:18]` instead of `[00:07]` and ended at `[18:14]` instead of `[18:18]`. The carry-over recovery also stamps the opening line 3 seconds late, since it arrives on the following cue.

## Approach

Rather than guess with fuzzy matching, this keys on the format's own structure. In rolling auto-subs only lines carrying inline word-timing tags (`<00:00:15.160><c> word</c>`) hold new content; carry-over lines never do. That identifies the real text exactly.

Detection is file-level, so **tracks without those tags — manual subtitles and Whisper output — keep their existing structural handling.** A word-overlap guard in `_dedupe` catches leftovers (including the untagged settle cues at the end of a track). It requires a 3-word run before trimming, so an incidental shared word between two cues is kept rather than risking dropped speech: a stray duplicate is cheap, lost words are not.

Cue bodies now terminate on the real blank separator and skip padding.

Also unescapes HTML entities, so a speaker change reads `>>` instead of `&gt;&gt;`. Happy to pull that out if you would rather keep the diff to the parsing fix.

## Result

Same talk, `scripts/watch.py --detail transcript`:

| | before | after |
|---|---|---|
| words | 5,975 | 3,035 |
| speaking rate | 323 wpm | 164 wpm |
| consecutive pairs sharing a 3+ word run | 465 | 0 |
| track span | `00:18` → `18:14` | `00:07` → `18:18` |

164 wpm is the plausible rate for a conference talk. Timestamps stay aligned to the words they introduce, and `--start` / `--end` filtering is unaffected.

## Tests

Adds `tests/test_transcribe.py` — 12 cases over a faithful rolling-sub fixture, covering dedup, the padding line, entity unescaping, timestamp alignment, the conservative short-overlap behavior, and a plain non-rolling track passing through untouched.

**7 of the 12 fail against the current parser.** Full suite goes 71 → 83 passing, nothing else changed.

I left the version alone and put the changelog entry under `[Unreleased]` — your call on release.


frinsen:
Thanks @eltokhy — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored — measured against a real rolling VTT: 9 056 → 4 552 words, matching your 2× finding.

PR #184, opened 2026-08-31
Add opt-in yt-dlp cookie support

endiaye677:
## Problem

Some sites refuse anonymous yt-dlp requests. YouTube in particular now answers ordinary public videos with:

```
ERROR: [youtube] <id>: Sign in to confirm you're not a bot.
Use --cookies-from-browser or --cookies for the authentication.
```

This fails at the download step, before any frame extraction or transcription runs — so the skill returns nothing at all. There was no way to supply cookies short of editing the script, since `watch.py` is the only entry point to yt-dlp.

## Change

A `_cookie_args()` helper reading two environment variables, applied to both yt-dlp invocations (`fetch_captions` and `download_url`):

| Variable | yt-dlp flag |
|---|---|
| `WATCH_COOKIES_FILE=/path/to/cookies.txt` | `--cookies` |
| `WATCH_COOKIES_FROM_BROWSER=chrome\|safari` | `--cookies-from-browser` |

Both unset by default, so the signed-out path is byte-for-byte unchanged.

## Docs

`SKILL.md` needed correcting, not just extending. The security section asserted:

> Does not access any platform account (no login, no session cookies, no posting) — yt-dlp only ever requests public data

That guarantee is untrue once a user opts in, and it is exactly the kind of claim someone reads before granting a tool access to their machine. This PR:

- Scopes that bullet to the default path and states what changes when cookies are enabled (read-only; never posts or mutates account state)
- Declares the capability under **What this skill does**, noting cookies go only to the URL's host and are never copied into the working directory or printed
- Points the download-failure mode at the flags, while instructing the model to **ask before reading a user's cookies** rather than reaching for them unprompted

## Testing

`main` is currently red from an unrelated ffmpeg 9 bug (21 of 71 tests fail; fixed in #183). Layering that fix in to get a green base, this branch is **71 passed** — the cookie change regresses nothing.

Not covered by tests: `tests/` has nothing referencing cookies, so the argv construction is unverified. Worth adding before merge if you'd like — happy to include it here.


frinsen:
Thanks @endiaye677 — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — `WATCH_COOKIES_FILE` / `WATCH_COOKIES_FROM_BROWSER` as you designed them, read through the shared `.env` parser and applied to all three yt-dlp invocations; your SKILL.md "ask before reading cookies" wording is in.

PR #187, opened 2026-09-02
Add `--lang` so non-English videos get their native captions

Rchardd:
## Problem

The caption language is hardcoded to English in two separate places:

- both yt-dlp invocations pass `--sub-langs "en.*"` (`fetch_captions`, `download_url`)
- `_pick_subtitle()` then additionally prefers `.en.` / `.en-US.` / `.en-GB.` / `.en-orig.` filenames

So a video whose captions exist only in another language comes back with **no transcript at all**. `/watch` then either:

- burns a paid Whisper call to re-derive captions the platform was already serving for free, or
- with no API key configured, degrades to frames-only and answers the user's question with none of the spoken content

I hit this on a Portuguese video that has perfectly good native `pt` captions. `yt-dlp` was asked for `en`, got nothing, and the report said `No transcript available`.

## Change

`--lang` takes a **comma-separated priority list**, default `en,pt`.

```bash
python3 watch.py "<url>"              # en, then pt
python3 watch.py "<url>" --lang pt    # pt only — one request
python3 watch.py "<url>" --lang pt,en # invert the priority
```

The first requested language that actually *produced a track* wins. That also handles yt-dlp failing one variant while another succeeds — common when a 429 kills one track mid-run, which I hit repeatedly while testing.

Region codes are normalised (`pt-BR` → `pt`) and duplicates collapsed.

> **Default choice is yours to make.** I used `en,pt` because it keeps English first for every existing user and `pt` is what I needed. If you'd rather not bless a second language in the default, `en` alone works identically for English users — the flag is the substance of the PR, not the default.

## The regression guard is preserved

`tests/test_download.py` guards against `--sub-langs all`, which makes yt-dlp enumerate hundreds of auto-translated tracks and stall before the video download starts. That guard stays.

What changed is the assertion: it previously required every token to start with `en`, which conflates "bounded" with "English". It now asserts the argv equals *exactly the caller's requested languages* — the property the guard was actually protecting, and a stricter check than before.

## Tests

Six new cases, all passing (`8 passed` in `test_download.py`):

- defaults resolve to `en,pt` in both yt-dlp call sites
- an explicit `--lang es` produces exactly `es.*`
- `_lang_list` normalises regions, dedupes, and falls back to `en` on empty input
- `_pick_subtitle` honours priority order when several tracks exist
- `_pick_subtitle` skips a requested language that produced no file
- `_pick_subtitle` still falls back to any available track, and returns `None` when there are none

No new failures in the rest of the suite — the pre-existing failures are unchanged in count and identity.

## Verified end-to-end

Against a Portuguese YouTube video that previously produced an empty transcript: **127 caption segments via captions**, with no flags passed.

## Note on stacking

This branch is cut from `main` and does **not** include the `-fps_mode` fix from the companion PR. On ffmpeg 8+ you'll want both — that one is required for any frame extraction to work at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable changes to `/watch` are documented here.
 ### Added
 - **`--lang CODES`** — comma-separated caption languages in priority order (default `en,pt`). The first requested language that yields a track wins; region codes are normalised (`pt-BR` → `pt`) and duplicates collapsed.
 
+### Changed
+- Captions are now requested as exact codes (`en,en-orig,pt,pt-orig`) instead of the previous `en.*` glob. yt-dlp matches `--sub-langs` case-insensitively, so a `code.*` glob also selects YouTube's auto-translated pairs (`pt-en`, `en-de`, `pt-PT-de`, …) — on one test video that was 7 caption fetches where 1 was wanted, and the surplus requests drew HTTP 429s that cost the transcript entirely.
+
 ### Fixed
 - Captions were requested as `en.*` only, hardcoded in both yt-dlp calls, and `_pick_subtitle()` further preferred `.en.`/`.en-US.`/`.en-GB.`/`.en-orig.` filenames. Any non-English video therefore came back with no transcript — falling through to Whisper (a paid API call) or, keyless, to frames-only, even when the platform had perfectly good native captions in the video's own language.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to `/watch` are documented here.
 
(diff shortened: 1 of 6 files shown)


frinsen:
Thanks @Rchardd — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — `--lang` is in, as a single preferred language that overrides detection (the default is now the video's own language, so the `en,pt` list wasn't needed); your point about `code.*` globs drawing 429s shaped the bounded `code,code-orig,code-XX` pattern.

PR #192, opened 2026-09-02
Fix UnicodeEncodeError crashing the report on Windows

IsaiahCalvo:
## The bug

On Windows, `/watch` crashes when it prints its report:

```
UnicodeEncodeError: 'charmap' codec can't encode character '→'
  File "skills/watch/scripts/watch.py", line 280, in main
```

The report contains two non-ASCII glyphs — an em dash in the work-dir footer and an arrow (`→`) in the focus-range line (`watch.py:281`) and the filtered-transcript headers (`watch.py:361,375`). Windows consoles default to cp1252, which can encode neither.

The painful part is *when* it fails: after the video has downloaded and the audio has already gone to Whisper. The user pays the full wall-clock and API cost, then gets a traceback instead of a report.

- Any run with `--start` / `--end` dies every time.
- Non-focused runs survive but print the footer as mojibake (`— delete when done` renders as `â€" delete when done`).

Reproduced on Windows 11 / Python 3.12.13 with a stock install.

## The fix

Reconfigure `stdout`/`stderr` to UTF-8 once at each entry point, rather than escaping the glyphs at each of the five print sites — so future non-ASCII in the report can't reintroduce the bug.

- `config.force_utf8_output()` — new shared helper next to the other config helpers.
- Called first thing in `watch.py:main()` and `setup.py:main()`.
- `errors="replace"` so a genuinely undecodable terminal degrades instead of crashing.
- `getattr(stream, "reconfigure", None)` guard, since pytest's capture swaps in objects without `reconfigure`.

No behavior change on macOS/Linux, where the streams are already UTF-8.

## Tests

Added `test_force_utf8_output_survives_cp1252_console`, which pins `sys.stdout`/`sys.stderr` to strict cp1252 wrappers, asserts the report line raises `UnicodeEncodeError` first, then calls the helper and asserts the same line writes with the arrow intact. It fails on `main` and passes with this change (confirmed by reverting just the `reconfigure` call).

Also added a lighter check that the helper leaves both real streams on UTF-8.

## Verification

`python -m pytest tests/ -q` → **70 passed, 3 failed**.

Those 3 failures (`test_keyless_completed_setup_proceeds_silently`, `test_dedup_collapses_static_by_default`, `test_no_dedup_preserves_static_frames`) **also fail on unmodified `main`** on this machine, so they're pre-existing and unrelated — I left them alone rather than widen the scope of this PR. Happy to look at them separately if useful.

End-to-end on Windows with `PYTHONIOENCODING` unset, which previously crashed:

```
- **Focus range:** 00:10 → 00:20 (10.0s)
- **Frames:** 2 selected from 3 candidates (uniform with uniform fallback, focused range, budget 2, cap 2)
- **Transcript:** 1 segments in range (via whisper (groq))
_Work dir: `...\watch-3_ko7yu3` — delete when done._
```

---

Unrelated, but worth a line in the README for other Windows users: `yt-dlp` needs both a JS runtime (`deno`) **and** `--remote-components ejs:github`. Without the solver script it silently selects a **video-only** stream, and Whisper then fails with "no audio stream found" — which reads like a Whisper problem but isn't. Also note `--js-runtimes deno:<path>` only parses with forward slashes on Windows; backslashes silently fail to register.


frinsen:
Thanks @IsaiahCalvo — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored — the `config.force_utf8_output()` approach with your cp1252 regression test.

PR #199, opened 2026-09-05
Add optional local/self-hosted Whisper backend

jayvee6:
## Why

Transcription is currently Groq or OpenAI only, so a caption-less video's audio has to leave the machine to get a transcript. That's a blocker for anyone running `/watch` against private local footage — the video never leaves (which the docs are rightly proud of), but the audio does.

This adds a third backend pointing at any server that exposes OpenAI's `/v1/audio/transcriptions` route — whisper.cpp's `server`, faster-whisper-server, speaches, LM Studio. Combined with a local file input (which already bypasses yt-dlp) and captions being absent, `/watch` becomes fully offline.

## How

Configured with `WATCH_WHISPER_BASE_URL`, plus optional `WATCH_WHISPER_MODEL` and `WATCH_WHISPER_API_KEY`. These are read with the same `env → ~/.config/watch/.env → ./.env` precedence as the existing keys, reusing `config.read_env_file` so quoting and inline-comment handling stay identical.

- `local_endpoint()` resolves the target. A bare origin (`http://localhost:8080`) gets the route appended; a full transcriptions URL is used verbatim.
- **When configured, local takes precedence over Groq/OpenAI.** The reasoning: setting the variable is itself the opt-in to keeping audio local, and it shouldn't be silently outranked by a key set months earlier. `--whisper groq|openai` still forces the remote path, and `--whisper local` is now a valid choice.
- The `Authorization` header is **omitted entirely** when no key is set. An empty `Bearer ` makes some local servers return 401 rather than treating auth as absent.
- `transcribe_video`'s key guard now permits an empty key for `local` only; Groq and OpenAI still require one.
- `setup.py`'s scaffolded `.env` documents the three new variables, commented out.

## Compatibility

No behavior change when `WATCH_WHISPER_BASE_URL` is unset — Groq is still preferred, OpenAI is still the fallback. All **71 existing tests pass unchanged**; 11 new ones bring the suite to 82.

The new tests live in `tests/test_local_whisper.py` rather than being folded into `test_whisper.py`, to keep the diff off your existing file. They cover endpoint resolution (bare origin, trailing slash, full URL), precedence (local wins by default, explicit `--whisper groq` still overrides, `--whisper local` without config returns nothing), dotenv reads, a regression guard that Groq→OpenAI ordering is untouched, and the conditional auth header in both directions.

## What I did not verify

I did not run this against a live local Whisper server — endpoint construction, header logic, and backend routing are unit-tested, but there's no end-to-end transcription test against a real instance. Worth a look from someone with one running before merge.


frinsen:
Thanks @jayvee6 — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored — one test adjusted because `.env` paths are now resolved per call via `Path.home()` (from #232), otherwise identical.

PR #200, opened 2026-09-05
fix(youtube): survive datacenter-IP bot gate; add storyboard frame fallback

drsandeeprana00-bit:
# fix(youtube): survive datacenter-IP bot gate; add storyboard frame fallback

## Problem

On datacenter IPs — hosted sandboxes (Claude's cloud container), CI runners, VPSes — YouTube refuses the default `web` player client with `HTTP 429: Too Many Requests` followed by `Sign in to confirm you're not a bot`.

`download.py` has no retry path, so:

- `fetch_captions()` returns `{'video_path': None, 'subtitle_path': None, 'info': {'url': ...}, 'downloaded': False}` — an empty result with no diagnostic
- `download_url()` raises `yt-dlp did not produce a video file`, which reads like a bad URL rather than an IP-reputation gate

Net effect: **every `/watch` on a YouTube URL fails on these hosts**, and the error message points the user in the wrong direction. Reproduced against `v0.2.0` on a Linux container:

```
WARNING: [youtube] Unable to download webpage: HTTP Error 429: Too Many Requests
WARNING: [youtube] Unable to fetch GVS PO Token for web client: Missing required Visitor Data
ERROR:   [youtube] Sign in to confirm you're not a bot.
RESULT: {'video_path': None, 'subtitle_path': None, 'info': {...}, 'downloaded': False}
```

Note this is **not** the TLS issue already handled by `--compat-options no-certifi`; that fix works correctly. This is a separate, later failure.

### 1. Player-client fallback (`download.py`)

`_run_ytdlp()` retries across `mweb`, `web_embedded`, `tv`, `android`, `ios` when the default client is refused. Gated on `_is_youtube()`, and **the default client is always attempted first**, so behaviour is unchanged on hosts that already work.

This alone is the high-value half. On a test video that previously yielded nothing, the fallback recovered full metadata and **304 caption segments** — meaning most affected users get a complete, correct result rather than a degraded one.

### 2. Storyboard fallback (`storyboard.py`, new)

When every client is refused for *media* but the page still resolves, YouTube generally still serves the storyboard: JPEG/WebP mosaics of thumbnails spanning the runtime. `storyboard.py` splits these into timestamped frames, so `/watch` degrades to low-resolution whole-video coverage instead of aborting.

`watch.py` surfaces this honestly in the report:

```
- Frames: 8 selected from 8 candidates (storyboard with uniform fallback, full range)
- Degraded mode: media download was refused, so frames came from the storyboard
  mosaic. Coverage spans the whole video but resolution is low (on-screen text is
  often unreadable) and there is no audio, so Whisper cannot run.
```

The explicit degraded-mode line matters: without it the model reads low-res tiles as if they were normal frames and over-trusts what it thinks it can see.

## Implementation notes

Three things that are easy to get wrong and are worth flagging for review:

- **Storyboard encoding varies by client.** The same video returns JPEG via one client and WebP via another. Checking a single magic number silently yields zero frames. Mosaics are identified by magic bytes across JPEG/WebP/PNG.
- **`email.message_from_binary_file` corrupts the mosaics.** The images are raw binary between MIME boundaries and the stdlib parser does line-based text handling, producing visibly mangled JPEGs that still decode. Parsing is done at the byte level instead.
- **Timestamps come from the mhtml figcaptions**, not from assuming uniform spacing. Grid dimensions are inferred from mosaic size, preferring standard YouTube tile widths.

## Dependencies

None added. `storyboard.py` is stdlib + `ffmpeg`, consistent with the rest of the package. Pillow is used as an optional fast path for near-duplicate dropping and is never required — without it, frames are sampled evenly.

## Testing

- Patch applies cleanly to a pristine `v0.2.0` tree (`git apply --check`, `git am`)
- All three modified/added files parse
- End-to-end `/watch` on a bot-gated host: previously total failure, now returns 304 caption segments plus timestamped storyboard frames
- Default-client path is untouched, so unaffected hosts are unaffected

**Not verified by me:** the happy path where the default client succeeds. I have no host that can reach YouTube unimpeded, so I could not confirm normal operation end-to-end. The fallback only engages after the default is tried and the local-file path is untouched, so the risk is structural rather than behavioural — but a maintainer should confirm on a normal machine before merging.

## Related

Users hitting this currently work around it with `--cookies-from-browser`, which is unavailable on headless hosts — the exact environments where this bites hardest.


frinsen:
Thanks @drsandeeprana00-bit — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — the `player_client` retry landed for the *media* fetch only (alternate clients drop caption tracks, and captions are fetched separately), capped at `mweb`/`tv`/`web_embedded`, triggered only on 403/429/bot-check output. The storyboard fallback needs Pillow so I left it out — happy to see it as a follow-up.

PR #204, opened 2026-09-06
Fix frame extraction on ffmpeg 8.x, which removed -vsync

mrrobotbuilder:
## The problem

ffmpeg deprecated `-vsync` in 5.0 and **removed** it in 8.0. `frames.py` passes it in two places, so on a current ffmpeg both engines abort before decoding anything:

```
Unrecognized option 'vsync'.
```

`extract_scene_candidates` and `extract_keyframes` raise `SystemExit`, which takes down **three of the four detail modes** — `efficient`, `balanced` (the default) and `token-burner`. Only `transcript` still works, so `/watch` returns no frames at all on ffmpeg 8.

It presents as a malformed command line rather than a version gap, which makes it easy to misdiagnose.

## The fix

A `vfr_flag()` helper that probes `ffmpeg -h full` once and returns `-fps_mode vfr` where supported, `-vsync vfr` otherwise. `-fps_mode` landed in 5.0, so this doesn't trade one version break for another — it keeps working on older builds too. Result is cached, so it costs one subprocess per process.

## Also included

`_frame_lines` in `tests/test_watch.py` matched the POSIX-only substring `"/frames/frame_"`. On Windows the report carries backslashes, so it counted zero frames and every frame-count assertion failed for a reason unrelated to what it was testing. Now matches either separator.

## Verification

On ffmpeg `N-125875` (8.x, Windows 11):

| | before | after |
|---|---|---|
| `tests/test_watch.py` | 8 failed, 2 passed | all passing |
| full suite | — | 71 passed, 1 failed |

The one remaining failure is pre-existing and unrelated: `test_setup.py::test_keyless_completed_setup_proceeds_silently` expects silence, but `setup.py` warns that `~/.config/watch/.env` is world-readable — Windows has no `chmod 600` equivalent, so it warns unconditionally. Left alone as out of scope for this PR; happy to open a separate one if useful.

Added `test_vfr_flag_is_accepted_by_the_installed_ffmpeg`, which asserts the probe result is valid and cached, then runs a real extraction through the flag so the regression can't come back silently.

https://claude.ai/code/session_01Y4qDCFb5PsrvNvFa1NoZ3Q


frinsen:
Thanks @mrrobotbuilder — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — the `OSError` guard from your `vfr_flag()` went into #219's probe; your `_frame_lines` separator fix landed via #208.

PR #205, opened 2026-09-06
Frame-coverage fixes and an on-device Whisper fallback

Jordan-Zhu:
Seven commits, each reviewable on its own. Three fix cases where the frame budget was spent in the wrong place; three add a keyless transcription path; one is a Windows crash.

### Frame coverage

**`_even_sample` spaced candidates by list index, not timestamp.** Every capped engine routes through it, so a cluster kept its proportional share of the budget: 36 keyframes packed into 6% of a video held ~90% of the slots and the other 94% of the runtime got what was left. `_even_time_indices` pins first and last, then gives each interior slot the candidate nearest its evenly-spaced target time. Degenerate timelines fall back to the old index spacing. Existing contracts unchanged — sorted, exactly `n`, reindexed `0..n-1`, spans first→last. The over-cap `--timestamps` cue path got the same treatment.

**`-vf fps=N` with `-frames:v max_frames` only covers `max_frames/N` seconds.** ffmpeg stops decoding at the cap, so any cap below `fps * duration` spent the whole budget at the head of the range: 4 frames at 2fps watched the first 2s of a 5.6s clip. Now fps is lowered to `max_frames/span`.

**`MAX_FPS` is overridable via `WATCH_MAX_FPS`.**

A related scene-detection coverage bug is in #210, separately.

### On-device Whisper fallback

A keyless install had no transcription path at all. Nothing about Whisper actually requires the network.

`local_whisper.py` transcribes on-device with faster-whisper (CTranslate2), `large-v3`, `device="auto"`, `compute_type="int8"` — no torch, no GPU, no key. It reuses `whisper.extract_audio` and returns `{start, end, text}` in the cloud path's shape, so `filter_range`/`format_transcript` are untouched. Lazy import. Precedence is either/or on whether a cloud key exists; `--no-whisper` disables both; audio never leaves the machine on the local path.

**`vad_filter` is deliberately off**, and it's the thing to review hardest, because I originally had it on and it was destroying transcripts. Full 114-minute film, `small` model, only the flag changed:

| | segments | last segment |
|---|---|---|
| `vad_filter=True` | 785 | 1:33:03 |
| `vad_filter=False` | **1543** | **1:53:44** |

VAD-on also dropped `1:33:02`–`1:43:16` outright, and the truncation point moved depending on whether the audio was passed whole or sliced, so the loss scales with input length.

**It is a trade, not a free win.** With VAD off, silence reaches the model and it confabulates. On that same film it invented a cleanup crew answering *"yes, ma'am"* over a scene the frames show as one person standing alone. Losing 21 real minutes is worse than gaining a few invented lines, so off stands — but the comment now says both halves.

**Low-confidence segments are marked, never dropped.** Whisper computes per-window doubt and we were discarding it; segments past its usual cutoffs now render `[?]`. Scope, measured rather than assumed: **this catches degraded audio, not invention.** I checked whether it would have caught the fabricated passage above — it would not. Those lines scored `no_speech_prob 0.04 / avg_logprob -0.35`, *more* confident than the real dialogue beside them, because the numbers describe decoding certainty rather than truth. They're also per decode window, not per segment. Both limits are in the code comment so nobody mistakes a clean transcript for a verified one.

### Windows

`watch.py` reconfigures stdout/stderr to UTF-8. Windows defaults to cp1252, so a run would extract every frame and then die with `UnicodeEncodeError` printing the report.

### Testing

Three tests added: `test_even_time_indices_spreads_over_clustered_candidates`, `test_extract_lowers_fps_so_cap_spans_the_range` (real fixture), and `test_format_transcript_marks_uncertain_segments`. Full suite on Windows: 3 failed / 71 passed, versus 3 failed / 68 passed on `main` — same three pre-existing failures, no regressions. #206 and #208 clear all three.

No unit test for the local whisper path: it needs `faster-whisper` installed and the module is a thin wrapper around one library call, so a mock would only assert I passed the arguments I passed. It was exercised end-to-end on a full-length feature instead, which is what surfaced the VAD problem and then the hallucination behind it.

### DIFF
--- a/skills/watch/SKILL.md
+++ b/skills/watch/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: watch
 version: "0.2.0"
-description: Watch a video (URL or local path). Downloads with yt-dlp, extracts auto-scaled frames with ffmpeg, pulls the transcript from captions (or Whisper API fallback), and hands the result to Claude so it can answer questions about what's in the video.
+description: Watch a video (URL or local path). Downloads with yt-dlp, extracts auto-scaled frames with ffmpeg, pulls the transcript from captions (then a Whisper fallback — cloud API if a key is set, else local faster-whisper), and hands the result to Claude so it can answer questions about what's in the video.
 argument-hint: "<video-url-or-path> [question]"
 allowed-tools: Bash, Read, AskUserQuestion
 homepage: https://github.com/bradautomates/claude-video
@@ -13,7 +13,7 @@ user-invocable: true
 
 # /watch
 
-You don't have a video input; this skill gives you one. A Python script gets captions first, optionally downloads the video, extracts frames as JPEGs (scene-aware, or fast keyframes at `efficient` detail), gets a timestamped transcript (native captions first, then Whisper API as fallback), and prints frame paths. You then `Read` each frame path to see the images and combine them with the transcript to answer the user.
+You don't have a video input; this skill gives you one. A Python script gets captions first, optionally downloads the video, extracts frames as JPEGs (scene-aware, or fast keyframes at `efficient` detail), gets a timestamped transcript (native captions first, then a Whisper fallback — cloud API if a key is set, otherwise on-device faster-whisper), and prints frame paths. You then `Read` each frame path to see the images and combine them with the transcript to answer the user.
 
 ## Resolve `SKILL_DIR` (do this before any command)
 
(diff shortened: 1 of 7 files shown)


frinsen:
Thanks @Jordan-Zhu — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): two of the seven commits cherry-picked: time-spaced `_even_time_indices` and `WATCH_MAX_FPS` (now validated instead of crashing on a bad value). The `-frames:v` fix landed via #226, UTF-8 via #192, and the faster-whisper commits were replaced by #199's dependency-free OpenAI-compatible local backend — hope that's a fair trade.

PR #206, opened 2026-09-06
Make Windows file-permission handling honest instead of a no-op

Jordan-Zhu:
Two commits, independent of #205. POSIX mode bits do not govern access on Windows and cannot be set from Python or Git Bash, but three places in the codebase acted as if they could.

### The session-start nag

`check-setup.sh` warns when `~/.config/watch/.env` isn't mode 600 and tells the user to run `chmod 600`. On Git Bash / MSYS / Cygwin both halves are broken: `stat -c '%a'` reports a synthesized `644` with no relation to the real NTFS ACL, and `chmod` is a silent no-op. Confirmed the round trip on Windows 11 / Git Bash — `chmod 600` then `stat` still reads `644`. Result: a `SessionStart` warning every single session, with a fix that provably cannot clear it.

### The same bug in Python

`setup.py` called `CONFIG_FILE.chmod(0o600)` after writing the secrets file, and `_check_file_permissions` warned when `st_mode & 0o044`. On Windows `os.chmod` only toggles the read-only attribute, and `st_mode` is synthesized from that same attribute — so the write was inert and the check always fired.

Verified rather than assumed: created two files in the same directory, one with `chmod(0o600)` and one without, and diffed `icacls`. Identical, both inheriting `SYSTEM` / `Administrators` / user.

### What Windows actually gives you

```
C:\Users\jorda\.config\watch   SYSTEM:(I)(F)  Administrators:(I)(F)  jorda:(I)(F)
```

The user, SYSTEM and Administrators — inherited from the user profile — and no other ordinary user. That is the normal posture for user secrets on Windows; an administrator can read any file regardless. So the practical guarantee `0600` was reaching for mostly holds, but the mechanism advertised isn't the one in play.

### The change

Restrict on POSIX, skip on Windows, and document what each platform actually gets. `README.md`, `SKILL.md` (both mentions) and `AGENTS.md` no longer claim a flat `0600`. The mode check still runs unchanged on `Linux` and `Darwin` — verified against `uname -s` values for all five platforms.

I stopped short of `icacls /inheritance:r /grant:r`, which would make literal owner-only true. It strips SYSTEM and Administrators, breaking backup and AV agents, and buys nothing against an attacker who is already an admin and can take ownership. The stopping point is named in a code comment so it reads as a decision rather than an oversight. Happy to add it if you'd rather have the strict version.

### Correction

An earlier revision of this description claimed owner-only was "the default for anything created under the user profile." That was wrong — the parent directories above show the default includes SYSTEM and Administrators. The one `.env` I first inspected had an explicit owner-only ACL set by something other than this tool, which is what misled me. The conclusion is unchanged; the reasoning is now accurate.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,4 +47,4 @@ bash skills/watch/scripts/build-skill.sh   # → dist/watch.skill
 
 - Keep the version in sync across `skills/watch/SKILL.md` (frontmatter), `.claude-plugin/plugin.json`, and `.codex-plugin/plugin.json` when cutting a release.
 - Releasing: tag `vX.Y.Z` and push the tag; `.github/workflows/release.yml` builds `dist/watch.skill` and attaches it to the GitHub release.
-- Never commit real API keys or `.env` contents; keys live in `~/.config/watch/.env` (mode `0600`) at runtime.
+- Never commit real API keys or `.env` contents; keys live in `~/.config/watch/.env` at runtime (mode `0600` on macOS/Linux; inherited user-profile ACL on Windows).
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ On the first `/watch` call, the skill runs `scripts/setup.py --check`. If `ffmpe
 - **macOS** — auto-runs `brew install ffmpeg yt-dlp`.
 - **Linux** — prints the exact `apt` / `dnf` / `pipx` commands.
 - **Windows** — prints the `winget` / `pip` commands.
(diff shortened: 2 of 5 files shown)


frinsen:
Thanks @Jordan-Zhu — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored (both the hook and `setup.py` halves, plus your docs wording).

PR #207, opened 2026-09-06
Add --force-whisper to bypass native captions

D0mD0mD0m:
## Problem

Platforms often hand back **machine-translated** auto-captions for non-English videos. Captions always beat Whisper in `watch.py`, so on those videos there was no way to get a transcript of what was actually said.

Concrete case — a German YouTube video (`R403PGBArDY`). yt-dlp returned English auto-captions with rolling duplicate lines and corrupted proper nouns:

| Source | Text at 11:49 |
|---|---|
| auto-captions | "I work a lot with **cloud computing**" |
| Whisper (groq) | "…weil ich eben sehr viel mit **Cloud** arbeite" |

Speaker names and product names degraded the same way ("Jurian Ivanov", "Brevus"), and every caption line repeated half of the previous one.

## Change

`--force-whisper` skips both caption-parsing paths, so the existing Whisper fallback takes over. No change to the Whisper logic itself — it already keys off "no transcript yet".

- Mutually exclusive with `--no-whisper` (guarded with `ap.error`).
- Combines with `--detail transcript`: since captions no longer short-circuit the download, that path fetches **audio only** — 6.5 MB instead of a 35 MB video on the example above.
- Documented in `SKILL.md`, including the privacy note: audio now also leaves the machine when captions are deliberately bypassed.

## Testing

- New hermetic test for the mutual-exclusion guard (no network, in the style of the existing suite).
- Full suite: **72 passed**.
- Verified end-to-end against a captioned URL: without the flag `6 segments (via captions)`, with it `176 segments (via whisper (groq))` on the German video.

Version intentionally left unbumped — `AGENTS.md` puts version sync in the maintainer's release step.


frinsen:
Thanks @D0mD0mD0m — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — `--force-whisper` landed as you wrote it (mutually exclusive with `--no-whisper`), wired into the new caption flow.

PR #209, opened 2026-09-07
Fix frame extraction on ffmpeg 8.0+ (-vsync was removed)

erosvellucci:
## The bug

ffmpeg 8.0 removed `-vsync` (deprecated since 5.1 in favour of `-fps_mode`). `extract_scene_candidates` and `extract_keyframes` both pass `-vsync vfr`, so on ffmpeg 8+ every frame mode aborts before decoding a single frame:

```
[watch] extracting scene-aware frames over full 19.0s (target 19, cap 100)…
ffmpeg scene extraction failed: Unrecognized option 'vsync'.
```

Reproduced on ffmpeg 9.0.1 (current Homebrew). `--detail transcript` is unaffected — it never shells out to ffmpeg — so the failure only surfaces once a run actually needs frames, which made it look like a download problem at first.

The existing suite is already red on ffmpeg 8+: **9 failures in `tests/test_frames.py`** on `main`, all of them this.

## Why not just swap the flag

Replacing `-vsync` with `-fps_mode` breaks the opposite end. ffmpeg 4.x only understands `-vsync`, and it is still widely deployed — Ubuntu 22.04 ships 4.4, Debian bullseye 4.3 — both of which `setup.py` happily installs via `apt`. Neither spelling is safe to hardcode.

## The fix

`_frame_rate_mode_flag()` probes a null encode once per process and caches the answer with `functools.lru_cache`:

```python
ffmpeg -hide_banner -loglevel error -f lavfi -i nullsrc=s=16x16:d=0.04 \
       -fps_mode vfr -f null -
```

Exit 0 → `-fps_mode`, anything else → `-vsync`. The probe decodes one 16×16 synthetic frame and measured **~35 ms**, paid once, on a path that already spends seconds in ffmpeg. It fails open to `-vsync` on `OSError`, since the callers' own `shutil.which("ffmpeg")` guard reports a missing binary with a much better message than a probe failure would.

Kept in `frames.py` next to its callers rather than `config.py` — it is an ffmpeg capability, not user configuration.

## Tests

Three cases in `tests/test_frames.py`:

- `test_frame_rate_mode_flag_is_accepted_by_this_ffmpeg` — the real regression guard: whatever flag the probe picks must actually be accepted by the installed ffmpeg. This fails on `main` under ffmpeg 8+ and passes here, on any version.
- `test_frame_rate_mode_flag_falls_back_to_vsync` — non-zero probe (the 4.x case) selects the legacy spelling.
- `test_frame_rate_mode_flag_fails_open_when_ffmpeg_missing` — `OSError` does not propagate.

```
before:  9 failed in tests/test_frames.py   (ffmpeg 9.0.1)
after:  74 passed in 8.93s                  (full suite)
```

Also verified end to end outside pytest: `/watch` against a captioned YouTube URL at `--detail balanced` now extracts and writes frames, and the JPEGs read back as images.


erosvellucci:
Closing — this was only ever needed as a local fix on our side, not something we're looking to land upstream. Apologies for the noise.

PR #210, opened 2026-09-07
Fall back to uniform when scene changes cluster, not just when they are few

Jordan-Zhu:
One commit, independent of #205, #206 and #208.

`SCENE_MIN_FRAMES` only asks whether *enough* scene changes were found, never whether they landed anywhere useful. Focusing on the last ten minutes of a feature returned 23 candidates — comfortably over the floor of 8 — with 15 of them inside seven seconds and a 3.4-minute hole in the middle:

```
1:44:39 … 1:44:46   15 frames in 7 seconds
1:46:11             ← last frame before the gap
1:49:36             ← next frame, 3m25s later
```

The frames spanned the range and showed almost none of it, and the engine reported success. Because the count cleared the floor, no fallback fired.

`_covers_range` measures clustering instead. Spacing the candidates you actually have evenly would put one every `span / (n + 1)` seconds; a hole more than `MAX_GAP_RATIO` (4×) that means the frames skip a stretch of the video.

Worth flagging for review, since it's the non-obvious part: judging against the frame *budget* was my first attempt and it's wrong. It demands scene detection return as many cuts as the cap, which fails the 5.6s `cut_clip` fixture immediately and amounts to a slower way of never using the scene engine at all. The denominator has to be the candidate count.

Same range after the change: **98 frames evenly spaced every 6 seconds** instead of 23 bunched into one corner. The report also names the reason it fell back now — `uniform with uniform fallback` told nobody anything.

### Testing

`test_covers_range_rejects_clustered_candidates` uses the real observed shape above, plus a well-spread control and a case that clears the count floor while still skipping most of the range. `test_worst_gap_counts_the_edges` covers the helper, including the leading and trailing gaps.

Full suite on Windows: 3 failed / 70 passed, versus 3 failed / 68 passed on `main` — same three pre-existing failures, no regressions. #206 and #208 clear all three.

### DIFF
--- a/skills/watch/scripts/frames.py
+++ b/skills/watch/scripts/frames.py
@@ -24,6 +24,12 @@ SCENE_THRESHOLD = 0.20
 # this is a low floor — NOT the frame budget — so normal videos with cuts use
 # the (single-pass) scene engine instead of paying for a wasted second decode.
 SCENE_MIN_FRAMES = 8
+# ponytail: clearing SCENE_MIN_FRAMES is not the same as covering the range. A
+# 10-minute stretch of a film returned 23 scene candidates — enough to pass the
+# floor — with 15 of them inside seven seconds and a 3.5-minute hole in the middle,
+# so the frames "spanned" the range while showing none of it. Count can't see a
+# hole; measure the worst one against the spacing the frame budget implies.
+MAX_GAP_RATIO = 4.0
 # Below this many decoded keyframes a clip is too sparse for keyframe coverage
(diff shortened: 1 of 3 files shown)


frinsen:
Thanks @Jordan-Zhu — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored; the report line now combines your `fallback_reason` with #224's candidate count.

PR #212, opened 2026-09-08
Prefer the video's own caption language over English

AlexLenovo:
## The problem

`download.py` asks yt-dlp for `--sub-langs en.*` in both download paths. On a non-English video that does not fail — YouTube serves an English track anyway, machine-translated from the source language.

The result is a transcript that has been through two lossy steps instead of one: auto-transcription, then auto-translation. The errors of the first step are carried into the second, and nothing in the report says the transcript is a translation.

## Evidence

Measured on a German video (`BSyes3ApU8c`, 29:08):

| Track | Segments | Renders "AGI" as |
|---|---|---|
| `en` (auto-translated) | 722 | "AGE" |
| `de-orig` (the original) | 751 | "AGE" |

Both carry the same mis-hearing, which shows the English text is derived from the German auto-captions rather than from the audio. The native track also carries 29 more segments.

## The change

`_native_lang()` reads the `language` field from the info JSON yt-dlp already writes, so no extra request is needed to find it. When the video is not in English, `_subtitle_with_fallback()` fetches the video's own captions and prefers them. English remains the fallback when no native track exists.

- **English videos: unchanged, and no extra request.** `_native_lang()` returns `None` for `en` and `en-US`, and for a missing, null, or unparseable field.
- **At most one extra subtitle fetch**, only for non-English videos.
- Both download paths (`fetch_captions` and the video download) go through the same helper.

## Testing

Re-ran the same German video: the fallback fires, `video.de-orig.vtt` is selected, 751 segments.

`_native_lang()` checked against `en`, `en-US`, `de`, a missing `language` key, a null value, an absent info file and a malformed one — only `de` triggers the extra fetch.


frinsen:
Thanks @AlexLenovo — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — your `_native_lang()` idea (read `language`/`original_language`, fetch the native track even when English *did* come back) is the core of the new caption logic, combined with #221's ranking and #234's unknown-language fallback; `--lang` from #187 overrides it.

PR #214, opened 2026-09-08
Fix: don't send one provider's API key to the other provider's endpoint

nbkwabi:
## The problem

`transcribe_video()` calls `load_api_key()` with no argument, even when the caller has already forced a backend:

```python
if backend is None or api_key is None:
    detected_backend, detected_key = load_api_key()   # <- no `preferred`
    backend = backend or detected_backend
    api_key = api_key or detected_key
```

So with only `GROQ_API_KEY` set:

```bash
python3 watch.py video.mp4 --whisper openai
```

`backend or detected_backend` keeps `"openai"`, while `api_key or detected_key` picks up the **Groq** key — and `_transcribe_file` then POSTs that Groq key to `api.openai.com`.

This contradicts the guarantee in `SKILL.md`:

> Does not share API keys between providers (Groq key only goes to `api.groq.com`, OpenAI key only goes to `api.openai.com`)

It fails silently from the user's side: they get an auth error from OpenAI, with no indication their Groq key just left the machine for a third party.

## The fix

`load_api_key` already supports `preferred=` — it just wasn't being passed. One line.

Behaviour with no forced backend is **unchanged**: still prefers Groq, still falls back to OpenAI. When a backend *is* forced and only the other provider's key exists, it now returns `(None, None)`, so the user gets the existing "no API key available" message instead of a silent cross-provider send.

## Tests

Adds `TestBackendKeyIsolation` to `tests/test_whisper.py` — both directions, the unchanged no-preference fallback, and the specific regression. The fixture clears both env vars and repoints `HOME` and the cwd at a `tmp_path` so a real `~/.config/watch/.env` can't leak into the run.

I couldn't get `pytest` installed on my machine (broken `ensurepip`), so I verified the same five assertions with a standalone harness against the patched module — all pass. Worth a CI run to confirm the pytest form.

---

Thanks for the plugin — genuinely useful. Found this while reading through `whisper.py` before setting it up.


robinjose911:
Thanks for this — I hit the same bug and can confirm it's live: with `GROQ_API_KEY` empty and `OPENAI_API_KEY` set, `--whisper groq` builds a POST to `api.groq.com` carrying the OpenAI key. `SKILL.md` promises key isolation, and the troubleshooting section's "retry with the other backend" advice leads straight into it.

Since this repo has been inactive since June, I've cherry-picked your commit into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `fa239a1`) with your authorship preserved — full credit is yours, and it's recorded in CREDITS.md.

One note in case it's useful upstream: the tests in this PR pass against unpatched source, since they only exercise `load_api_key()`, which already honoured `preferred`. I added a test that drives `transcribe_video` and asserts on the request that would be sent, which does fail without the fix.

Would still much rather see this merged here.

frinsen:
Thanks @nbkwabi — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored, with your `TestBackendKeyIsolation` suite.

PR #218, opened 2026-09-12
Fix quoted .env values with trailing comments and duplicate parsers

jvdurian-pixel:
## Summary
- `read_env_file()` checked for quotes before stripping an inline comment, so a quoted value with a trailing comment (`WATCH_DETAIL="balanced"  # note`) never got unquoted and silently fell back to the default.
- `setup.py` and `whisper.py` each carried their own copy of this parser (used for `GROQ_API_KEY`/`OPENAI_API_KEY`), missing the comment-stripping fix entirely. Both now call `config.read_env_file()` instead of duplicating the logic.

## Test plan
- [x] `pytest tests/test_config.py -q` — 7/7 passing, including two new regression tests for the quoted+comment and unquoted+comment cases
- [x] Confirmed via `git stash` that the pre-existing `test_setup.py`/`test_whisper.py` failures on this machine (missing ffmpeg/yt-dlp) are unrelated to this change

https://claude.ai/code/session_015FMj1tVT4PQAk2kGxqw6Wq


frinsen:
Thanks @jvdurian-pixel — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — `_parse_value` (quotes first, then trailing comment) went in verbatim on top of #232's parser consolidation, with your two regression tests.

PR #219, opened 2026-09-12
fix(frames): ffmpeg 8+ compatibility — -vsync was removed

bagalutenGregor:
## Problem

On ffmpeg 8.0 or newer, every frame mode of `/watch` fails. `frames.py` passes `-vsync vfr` in both the scene pass (`extract_scene_candidates`) and the keyframe pass (`extract_keyframes`), and ffmpeg 8.0 removed the option after a long deprecation:

```
$ python3 scripts/watch.py clip.mp4 --detail efficient --no-whisper
[watch] extracting keyframes over full 6.0s (target 12, cap 50)…
ffmpeg keyframe extraction failed: Unrecognized option 'vsync'.
```

Same failure in `balanced` and `token-burner`. `transcript` is unaffected (no ffmpeg frame pass). Reproduced on ffmpeg 8.1.1 (Fedora 44) and 9.0.1 (Arch/Manjaro); ffmpeg 5.1.9 (Debian 12) is fine.

## Fix

Probe `ffmpeg -h full` once, cached via `functools.lru_cache`, and emit the flag that build understands:

- `-fps_mode vfr` when supported (ffmpeg 5.0+)
- `-vsync vfr` otherwise, so old builds keep working

Both call sites spread the result, so the two passes can never drift apart again. No behavior change on supported builds — `-fps_mode vfr` is the documented replacement for `-vsync vfr`.

## Tests

Three added to `tests/test_frames.py`:

- probe output containing `-fps_mode` selects `-fps_mode vfr`
- probe output with only `-vsync` falls back to `-vsync vfr`
- a live check that runs the *installed* ffmpeg with the flag the probe picked and asserts it exits 0 — this is the one that would have caught the original break in CI on a modern ffmpeg

Full suite: 74 passed (ffmpeg 9.0.1). A cache-clearing fixture keeps the faked probe from leaking into the tests that shell out to real ffmpeg.

## Verified

Both frame modes produce frames again on ffmpeg 9.0.1, 8.1.1 and 5.1.9 — the 5.1.9 box exercises the `-vsync` fallback path.


frinsen:
Thanks @bagalutenGregor — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored — your probe + tests are the fix that closed the 15 `-vsync` issues; I only added an `OSError` guard for a missing ffmpeg (from #204).

PR #220, opened 2026-09-13
feat(download): Support for TikTok photo slideshows

Rasmus257:
## Why

`/watch` can't read TikTok photo slideshows, and the slides are the whole content of those posts. yt-dlp rejects `/photo/` URLs as `Unsupported URL`. On `/video/` URLs it downloads only the soundtrack mp3, and frame extraction then crashes with `ffmpeg scene extraction failed: Input #0, mp3`.

## What changed

- **gallery-dl fallback.** When yt-dlp returns no video, or only audio when frames were wanted, `download.py` falls back to `gallery-dl` if it's on PATH. It returns the slide images, soundtrack and caption, or the video file for posts yt-dlp couldn't fetch. It's optional: without gallery-dl, behavior matches today apart from the crash fix below.
- **Slides in the report.** Slides are scaled with the existing 512px/1998px clamp and are not deduped. The report lists them as `slide N`, with the post caption as Title. A 9-slide post is about 4k image tokens at the default width.
- **No crash on audio-only downloads.** A download with no video stream now skips frame extraction instead of crashing, and still goes on to Whisper.
- **Tests.** 9 new tests with no network: fallback routing, slide extraction, an in-process report run, and the audio-only guard. 6 of them fail on `main` and pass here; the full suite passes (80). Also checked by hand on real slideshow and video URLs, with and without yt-dlp browser impersonation.


frinsen:
Thanks @Rasmus257 — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted in part — the crash when yt-dlp hands back only the mp3 is fixed (the run finishes transcript-only with a clear Frames line). The gallery-dl slideshow support is deferred, listed in the CHANGELOG as a candidate.

Issue #107, opened 2026-07-29
Windows: secrets permission check always warns (st_mode does not reflect NTFS ACLs)


Elenics:
### Summary

On Windows the secrets-permission check can never pass, so every session prints a warning the user has no way to clear.

### Environment

- Windows 10, installed as a Claude Code plugin (`watch@claude-video` 0.2.0)
- Native Windows Python 3.12 (not WSL); Git Bash runs the SessionStart hook
- Setup is otherwise healthy: `setup.py --json` reports `"status": "ready"`, `"can_proceed": true`

### What happens

Both the hook and `setup.py` warn on every run:

```
/watch: WARNING — /c/Users/natio/.config/watch/.env has permissions 644 (should be 600).
  Fix: chmod 600 /c/Users/natio/.config/watch/.env
```

```
[watch] WARNING: C:\Users\natio\.config\watch\.env is readable by other users. Run: chmod 600 C:\Users\natio\.config\watch\.env
```

### Why the user can't fix it

On Windows, `os.stat().st_mode` is not derived from the NTFS ACL — CPython synthesizes it from the read-only attribute, so an ordinary file always reports `0o644`/`0o666`. MSYS/Git Bash `stat -c '%a'` reports the same. Consequences:

- `chmod 600` is a no-op — verified: `before=644` → `chmod 600` → `after=644`
- Even after locking the file down to the owner account only, both checks still report 644:
  ```
  icacls .env /inheritance:r /grant:r "natio:(F)"
  ```

So the file genuinely *is* private, but neither check can observe that. The warning is unavoidable on every Windows install.

Related: `CONFIG_FILE.chmod(0o600)` (`setup.py:136`, `setup.py:161`) also silently does nothing on Windows, so a fresh Windows install starts in the "failing" state by construction.

### Suggested fix

Skip the mode check where it can't carry meaning:

- `skills/watch/scripts/setup.py:71` — early-return from `_check_file_permissions()` when `platform.system() == "Windows"`
- `hooks/scripts/check-setup.sh` — guard the `stat` block, e.g. `[[ "$OSTYPE" != msys* && "$OSTYPE" != cygwin* && "$OSTYPE" != win32 ]]`

If you'd rather keep real coverage on Windows, the equivalent signal is the ACL (`icacls` output, or `win32security`) rather than the mode bits.

Happy to send a PR if that would help.

Jochen-1979:
[...]

PR #221, opened 2026-09-14
Prefer human-authored captions over auto-generated ones

oheewono:
## Problem

When a video has **both** a human-authored caption track and auto-generated
ones, `/watch` picks the auto-generated one.

`_pick_subtitle()` ranks by filename sort, and `video.en.vtt` sorts before
`video.en-US.vtt`. yt-dlp is asked for `--sub-langs "en.*"` with both
`--write-subs` and `--write-auto-subs`, so all the variants land in the work
dir and the auto track wins on every video where the two coexist. The module
docstring already promises the opposite:

```python
Also fetches subtitles (manual first, then auto-generated) in VTT format so
```

The filename can't distinguish them — manual and auto are both written as
`video.<lang>.vtt`. But yt-dlp does record it: `info.json` lists human tracks
under `subtitles` and machine ones under `automatic_captions`.

## Why it matters

YouTube's auto-captions are *rolling* — each cue repeats the tail of the
previous one. Real output from `NASA JPL "Mission Overview: Perseverance"`
(`5qqsMjy8Rx0`), which has a manual `en-US` track available:

**Before** — auto `en` track:

```
[00:10] good chance of having ancient life the perseverance rover will land at a
[00:12] the perseverance rover will land at a location called jezreel crater
[00:14] location called jezreel crater jezreel crater is a very interesting
```

**After** — manual `en-US` track:

```
[00:10] - [Man] The Perseverance Rover will land
[00:12] at a location called Jezero Crater.
[00:14] Jezero Crater is a very interesting place.
```

Three separate costs:

- **Roughly 2.6x the tokens.** 173 cues / 28 KB versus 67 cues / 8 KB for the
  same 2:58 video. Every duplicated half-line is transcript the model pays for
  twice.
- **Wrong text.** "Jezero Crater" comes back as "jezreel crater" throughout —
  the model then answers questions about a place that does not exist.
- **No punctuation or speaker labels.** The manual track marks `- [Woman]` /
  `- [Man]`, which matters on any multi-speaker video.

## Fix

Read the manual-track language codes from `info.json` and rank those first.
Falls back to the existing `en` > `en-US` > `en-GB` > `en-orig` order when
nothing is manual, so videos with only auto-captions behave exactly as before.

## Verification

Environment: Windows 11, ffmpeg 9.0.1, yt-dlp 2026.08.19, Python 3.13.

**Picker, against a real work dir** with all three tracks present:

```
manual subtitle langs: {'en-US'}
selected: video.en-US.vtt     (was: video.en-orig.vtt)
```

**End to end** on the video above — the after-excerpt is the actual output.

**Test suite** — no regression:

| | this branch | main |
|---|---|---|
| `pytest -q` | 24 failed, 47 passed | 24 failed, 47 passed |

With #216 and #217 also applied: **3 failed, 68 passed** — the same 3 as on
those PRs (#107's ACL warning, and two static-clip dedup tests I flagged in
#217 and have not attributed).


robinjose911:
Thanks for this — the human-authored preference is the right call and I've taken it as-is.

One change I'd suggest, and did make downstream: the ranking among **auto** tracks puts a bare `en` ahead of `en-orig`. On a non-English video, `en` is YouTube's machine translation while `en-orig` is the source-language track, so this silently swaps an original transcript for a translated one. That's the complaint in #123/#144/#153, arriving through a different door. I reversed it to `en-orig` first.

Worth noting the previous behaviour was accidentally correct here: `sorted()` puts `video.en-orig.vtt` before `video.en.vtt` (`-` sorts before `.`), so the old `preferred[0]` picked `en-orig`. The PR makes that ordering explicit — just the other way round.

Verified the change is free where it doesn't apply: on an English video the two tracks are byte-identical (107504 bytes each).

Also added five tests, since the PR ships without any — manual-beats-auto, manual English, the orig-over-translated ordering, no tracks, and a missing info.json. The ordering test fails against the original ranking.

Cherry-picked into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `eba9ea8`, with my follow-up in `cb7081f`), authorship preserved and credited in CREDITS.md. Would still rather see this merged here.

frinsen:
Thanks @oheewono — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — the human-vs-auto ranking from `info.json`'s `subtitles`/`automatic_captions` is the top rule of the new caption picker, combined with #212/#234's native-language fetch.

PR #224, opened 2026-09-16
Fix uniform fallback reporting more frames selected than candidates

dsp407:
Both engines fall back to the uniform sampler when their detector comes up short, but reported the rejected candidates as candidate_count:

  extract_scene_or_uniform  -> candidate_count=scene_count, the scene cuts
                               it just rejected, which never reach dedup or
                               selection
  extract_keyframes         -> candidate_count=len(candidates), keyframes
                               whose files it unlinks a few lines earlier

So the "- **Frames:**" line could report more selected than candidates, or arithmetic that cannot hold:

  12 selected from 3 candidates (uniform with uniform fallback, ...)
  3 selected from 3 candidates (..., 9 near-duplicates dropped, ...)

Both now report len() of the uniform frames actually extracted, captured before dedup, so candidates - dropped == selected holds.

On a fallback the engine is already "uniform", so " with uniform fallback" repeated the word while naming neither detector. A fallback_from key ("scene" / "keyframe") on both fallback dicts drives " after too few scene candidates" / " after too few keyframe candidates" instead.

Nothing covered the fallback metadata, so this changed candidate_count on both branches without breaking any of the 71 existing tests. Six tests now assert the accounting invariant (with and without dedup), the fallback_from key, and the wording and arithmetic of the rendered line. All six fail before this change.

Claude-Session: https://claude.ai/code/session_01A5oiyyQQmN5WQJqLpZcN7n


frinsen:
Thanks @dsp407 — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored.

PR #225, opened 2026-09-17
Fix YouTube rolling auto-captions shipping every line twice

OpenClawLinda:
## The bug

`transcribe._dedupe` collapses a cue only when it is byte-identical to the previous one, or a strict prefix extension of it. YouTube auto-subs are neither: they **scroll two lines per cue**, so line 2 of cue N reappears as line 1 of cue N+1. Neither branch ever fires, and the entire transcript ships duplicated.

Measured on a 27-minute talk with auto-generated captions: **64,511 bytes of transcript where the spoken content is ~28,500** — roughly 2.2x the transcript tokens on every YouTube video, and text that reads like a skipping record, with each line repeating the back half of the one before it.

Reproducible without the network via the fixture builder this PR adds:

```python
build_rolling_vtt(path, ["line one", "line two", "line three"])
# before: every sentence appears twice, in two adjacent cues
```

## The fix

A tail/head word-overlap check: find the longest run where the tail of the previous segment equals the head of the incoming cue, and strip those words off the **front** of the incoming cue.

Two details are load-bearing, and I got both wrong on the first attempt:

**1. The cue keeps its own start/end and stays its own segment.** Splicing it into the predecessor also deduplicates — and collapses a 27-minute track into **2 segments**. That leaves two `[MM:SS]` stamps for the whole video (so nothing to align frames against, which `SKILL.md` Step 3 depends on) and makes `filter_range()` return the entire transcript for *every* `--start/--end` window, silently turning focused mode into a no-op.

**2. The overlap check runs *before* the strict-prefix branch.** Stripping the overlap leaves a short segment that the next cue legitimately starts with, so a prefix check running first re-merges the pair and quietly undoes the granularity. The gap guard keeps the two branches apart: only back-to-back cues reach the overlap path, so genuinely growing cues still fall through to the prefix branch.

Guards against over-merging: an overlap must be **≥3 words** (shorter is coincidence, not a scroll) and the cues must be **within 1s** of each other — so a phrase the speaker genuinely repeats later in the video is never touched.

## Also fixed

The cue-reading loop used `lines[i].strip()` as the end-of-cue test, so a cue whose **first line is a lone space** — a placeholder YouTube emits — was silently dropped. On the same 27-minute video this ate the opening line. A VTT cue ends at a truly empty line.

## Tests

`tests/test_captions.py`, 5 cases, plus a `build_rolling_vtt` helper in `conftest.py`. The fixture is **synthesized**, matching the existing approach — network-free, and nothing third-party redistributed into the repo.

Each test is mutation-checked against the specific failure it exists for, rather than assumed to discriminate:

| Mutant | Goes red |
|---|---|
| current `main` (no overlap dedupe) | `test_rolling_cues_are_deduped`, `test_cue_starting_with_a_lone_space_is_not_dropped` |
| splice-style fix (dedupes correctly, merges cues) | `test_dedupe_preserves_segment_granularity`, `test_filter_range_returns_a_slice_not_everything` |

The second mutant is the one worth keeping: it deduplicates perfectly and passes every assertion aimed at the original bug, while breaking focused mode. That was my actual first implementation.

`filter_range` is asserted semantically (content from the end of the track must not surface in an early window) rather than as a size ratio — `filter_range` keeps any segment *overlapping* the window, so a character-count ratio on a short track proves nothing in either direction.

**Full suite: 76 passed** (71 existing + 5 new).

## Not included

While working through this I hit four other issues in the same skill, each a separate concern and not mixed in here: a bare `HTTP 403` on a stale yt-dlp (with no hint that the binary is the cause), no transcript-only fallback when the video stream fails but captions succeed, no staleness signal in preflight, and `-vf fps=N` + `-frames:v N` in the uniform `extract()` path returning only the head of the range — the same shape already documented for the scene path at `frames.py:529`. Happy to send any of them as follow-ups if useful.


robinjose911:
Thanks for this — it's a genuinely good fix and the reasoning in the comments saved me working out the ordering constraint myself.

Data point in case it's useful: I validated it against two real YouTube auto-caption tracks. It removed 49.8% and 49.9% of words respectively, and the output matched a dedupe I'd built by hand from the same files at 100% word-sequence similarity with identical word counts — so it's stripping duplication and nothing else. Segment granularity held (564 → 564 on one file), so focus mode still works.

Since this repo has been inactive since June, I've cherry-picked it into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `c5a9e57`) with your authorship preserved, credited in CREDITS.md.

Two small things: your commit email isn't linked to your GitHub account, so the commit shows your name without a profile link — you may want to add it to your account so this and #226 credit you properly. And one uncovered edge: a genuine stutter repeated within 1s across a cue boundary gets collapsed (`"well I think that" / "I think that is wrong"` → `"is wrong"`). Fine trade-off in my view, but might be worth a test documenting the intent.

frinsen:
Thanks @OpenClawLinda — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — #182 landed as the dedupe implementation (it uses the inline timing tags to identify new text); on a real 29-minute German auto-sub track both approaches produced the identical 4 552-word result, so yours validated it.

PR #226, opened 2026-09-17
Fix uniform extract() sampling the head instead of spreading across the range

OpenClawLinda:
## The bug

The uniform `extract()` builds its ffmpeg command with `-vf fps={fps}` plus `-frames:v {max_frames}`. The `fps` filter samples evenly across the whole input, but `-frames:v N` just makes ffmpeg **stop** after N output frames. So whenever the two disagree — `fps * duration > max_frames` — you get the first N samples, the head of the range, not a spread across it.

The auto path is safe by construction: `auto_fps`/`auto_fps_focus` derive `target` **from** `max_frames`, so the cap is never the binding constraint. The bug is reachable whenever fps and max_frames disagree, most obviously an explicit `--fps` override.

Measured on a 42s synthetic clip with `fps=2, max_frames=20`:

| | last frame timestamp |
|---|---|
| before | **4.5s** |
| after | near the end of the range |

On a real 27-minute video with `--fps 2`, that is the first ~50 seconds returned while the report says `100 frames, full mode` — it looks like a successful sample of the whole video and is a sample of the intro.

## This repo already knows this failure shape

`extract_scene_or_uniform`'s docstring says it outright — capping detection with `-frames:v` "would keep only the first `max_frames` cuts and drop the tail of long videos." The insight was applied to the scene engine and left in the uniform path. This PR just finishes the job.

## The fix

When `fps * range_duration` would exceed `max_frames`, lower the **effective** sampling fps so the capped frame count spreads across the requested range. `-frames:v` stays only as a rounding backstop for an odd boundary frame — it no longer decides coverage. `timestamp_seconds` is computed from the effective fps actually used, so reported timestamps match the frames that were really produced.

Scope is deliberately narrow: only the uniform `extract()`. The scene and keyframe engines are untouched.

## Tests

`tests/test_uniform_spread.py`, 5 cases, using an ffmpeg-synthesized clip via a `conftest.py` helper (network-free, consistent with the existing fixtures).

Proven red against current `main` before green — the two spread assertions fail with the head-only timestamps quoted above. Bounds were picked from what the broken code actually outputs, so correct and broken cannot both pass.

Two guards included deliberately:
- `test_scene_and_keyframe_engines_untouched` — this fix must not reach them.
- a snapshot of `auto_fps`/`auto_fps_focus` across 9+6 duration/cap combinations, pinning the auto budgets as unchanged. The fix is a no-op on that path by construction, and the snapshot keeps it that way.

**Full suite: 76 passed** (71 existing + 5 new), scene and keyframe coverage included.


robinjose911:
Thanks for this. Confirmed the bug on a 967.8s video — `extract(fps=2, max_frames=6)` returned six frames spanning t=0.0–2.5s, i.e. 2.5 seconds of a 16-minute video, reported as a full-range scan. After the fix the same call spans 806.5s.

Also checked the auto path you mention in the comment: with `fps` derived from `max_frames` the timestamps are byte-identical before and after, so the change is inert outside the conflicting case.

Cherry-picked into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `1f56858`) with your authorship preserved, credited in CREDITS.md, since this repo has been inactive since June. One note: it conflicts with your own #225 in `tests/conftest.py` — both append a helper, so it's a trivial keep-both, but worth knowing if they're ever merged in sequence.

frinsen:
Thanks @OpenClawLinda — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored, including the 42 s `long_cut_clip` fixture.

PR #227, opened 2026-09-17
Warn when yt-dlp is stale enough to be the cause of a silent 403

OpenClawLinda:
## The gap

`setup.py` checks dependencies with `shutil.which` — presence only. yt-dlp is the one dependency here that **rots**: YouTube rotates its client/signature scheme every few weeks, and a stale yt-dlp keeps existing, keeps being executable, and silently stops being able to download anything. `which` says yes forever.

Seen in the field: an installed yt-dlp ~76 days old made every download fail with a bare `HTTP Error 403: Forbidden`, while preflight reported everything fine and nothing pointed at the binary. Upgrading fixed it with no code change — same URL, same format selector, only the binary differed. The 403 reads like a blocked or region-locked video, so it sends you looking in exactly the wrong place.

## The fix

yt-dlp's version string **is** a release date (`2026.08.19`), so staleness is a pure local date comparison. Past 60 days, `--check` prints one warning line naming the age and the upgrade path.

**Threshold picked from data, not taste:** an observed healthy gap between consecutive releases was 46 days; the binary that actually started failing was ~76 days old. 60 sits above the healthy gap, so it won't nag right after an ordinary release cycle and train people to ignore it, while still giving roughly two weeks' warning before an age already proven to break downloads.

## What it deliberately does not do

- **No network call, ever.** Querying GitHub or PyPI for the latest release would add a failure mode and a privacy surface to something that runs on every invocation. The check only execs the already-installed binary.
- **Never changes the exit code.** 0/2/3/4 are byte-for-byte unchanged. Staleness is a warning printed on the ready path (still exit 0) and folded into the failure message otherwise. A stale binary that still works must not block a run.
- **Silent when fresh**, byte-for-byte — asserted directly, since silent-on-success is the contract `SKILL.md` tells the model to rely on.
- **Silent when unparseable.** Forks and distro builds use non-date version schemes, and a false "your yt-dlp is stale" on a good binary is worse than saying nothing. Invalid calendar dates (`2026.13.40`) are quiet too.
- **Skipped entirely when yt-dlp is missing** — that is already exit 2's job.

`yt_dlp_stale_days` is surfaced in `--json` so structured mode stays consistent with the human mode.

### On the lookup method

Shells out to `yt-dlp --version` rather than `import yt_dlp` and reading `__version__`. The import is faster, but it was verified not to work for a **pipx** install — this file's own recommended Linux path — where yt-dlp lives in an isolated venv the calling interpreter cannot import from. Only the binary's PATH entry resolves uniformly across pipx, brew, apt, pip --user and the standalone release binary.

`config.py` was left alone on purpose: it resolves user *configuration*, while this is a *binary preflight* concern that belongs beside the existing `_check_binaries()`.

## Tests

`tests/test_staleness.py`, 22 cases, network-free, with the version **injected** rather than read from whatever is installed — a test that only passes on today's machine is worthless next month.

Covers: stale warns with the right age; fresh is byte-for-byte silent; unparseable and invalid-calendar are silent; the exact threshold boundary and threshold+1; exit codes unchanged across every staleness outcome; `--json` consistency; and the check being skipped when yt-dlp is absent.

Mutation-checked rather than assumed: forcing the age computation to always report stale turns 16 cases red, and making staleness alter the exit code turns red exactly the two that name that contract.

**Full suite: 93 passed** (71 existing + 22 new).


frinsen:
Thanks @OpenClawLinda — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored; the staleness check now sits next to new impersonation / JS-runtime probes.

PR #228, opened 2026-09-17
Explain a 403, and keep the transcript when only the video stream fails

OpenClawLinda:
Two fixes to the same moment: what happens when yt-dlp cannot get the video.

## 1. A bare `HTTP 403` sends you looking in the wrong place

Seen in the field: a 27-minute video failed with `ERROR: unable to download video data: HTTP Error 403: Forbidden`, and the skill surfaced only `yt-dlp did not produce a video file in <dir> (exit 1)`.

The actual cause was a yt-dlp ~2.5 months stale that had lost YouTube's current client/signature rotation. Upgrading fixed it with no code change — same URL, same format selector, only the binary differed. But a 403 reads like a blocked or region-locked video, so the message points away from the real problem.

Now a 403/Forbidden failure says a stale yt-dlp is the likely cause, reports the installed version, and prints the upgrade command matching how yt-dlp was installed (pipx / brew / pip, with a generic fallback). **Every other failure keeps its existing message byte-for-byte.**

## 2. A dead video stream throws away a perfectly good transcript

In that same incident the subtitles downloaded fine — `video.en.vtt`, 272 KB — while only the media stream 403'd. The run raised `SystemExit` and discarded them. The transcript turned out to carry essentially all of the video's content.

The asymmetry is the bug: `watch.py` already handles *frames without transcript* gracefully everywhere, but had no mirror path for *transcript without frames*. For talking-head, lecture and podcast content, the transcript is most of the value.

Now that case completes in transcript-only mode instead of dying. Rendered output:

```
- **Video:** ⚠️ NOT OBTAINED — transcript-only mode, 0 frames (see warning below)
- **Frames:** 0 — NO VIDEO OBTAINED

> ⚠️ **NO VIDEO OBTAINED — TRANSCRIPT-ONLY MODE.** The video stream could not be
> downloaded, so there are **zero frames** and nothing in this run was watched
> visually. Answer using the transcript below only — do not describe or infer any
> visual content.
```

plus a replaced `## Frames` body and the same warning on stderr.

### Why it is loud rather than graceful

A report that looks normal while missing every frame is worse than the crash it replaces — a model reads this report and will otherwise answer as though it watched something. The degradation is stated in the body, not a footnote, and the warning explicitly instructs against inferring visual content.

### Deliberately kept distinct from `--detail transcript`

The underlying `video_path is None` plumbing is reused as-is — it already works regardless of *why* the path is null. The **wording** is not: `"skipped (transcript detail)"` describes a benign, user-chosen state, and reusing it for a genuine failure would under-communicate exactly what needs to be loud. A `degraded` flag, set only when a download was attempted and failed, layers the warnings on top of the shared data flow.

### Invariants

- Both missing (no video **and** no subtitle) still fails hard. A total failure never becomes a quiet success.
- `--start`/`--end` transcript filtering still works on the degraded path.
- Whisper is not attempted when there is no video to extract audio from.
- The sparse-long-video warning is suppressed when degraded — there are no frames for it to be about.
- A successful download is completely unaffected.

## Tests

`tests/test_download_failure.py`, 13 cases, using a **synthesized** stub `yt-dlp` added to `conftest.py` that reproduces the incident shape exactly: `--skip-download` succeeds and writes captions, the media download 403s. No network, and no real caption track or media file enters the repo.

Proven red against current `main` first: **10 of 13 fail**. The 3 that pass unfixed are regression guards rather than bug-discriminators, and are meant to pass in both states — non-403 messages unchanged, successful downloads unaffected, and both-missing still failing hard.

**Full suite: 84 passed** (71 existing + 13 new).


robinjose911:
Thanks for this — the asymmetry you identify is the real bug. `watch.py` handled frames-without-transcript everywhere and had no mirror path, and "loud rather than graceful" is the right call: a report that looks normal while missing every frame is worse than the crash.

That's your third fix I've ported. #225 and #226 are both in as well.

One bug worth fixing upstream: `_update_hint()` matches against `shutil.which()`'s result, but **uv and pipx both symlink into `~/.local/bin`** and keep the real venv elsewhere — so that string identifies neither.

```
which:     ~/.local/bin/yt-dlp
resolved:  ~/.local/share/uv/tools/yt-dlp/bin/yt-dlp
```

My uv-managed install gets told to run `yt-dlp -U`, which can't update a managed venv, and pipx falls through its own check. Resolving the link first and adding a `uv/tools` case fixes both. Given a wrong command lands exactly when someone is already stuck, it seemed worth being precise about.

Related: `test_403_failure_names_staleness_and_upgrade_command` calls the real `_update_hint()` and asserts against three hard-coded strings, so it passes or fails on how the developer installed yt-dlp — it fails outright on a uv install. Pinning the hint and checking the message carries it makes it host-independent, which is the same concern as #231.

Ported into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `4d5c5fd`, fix in `ad19640`) with your authorship preserved, credited in CREDITS.md. It needed a three-way merge against #235's cache and #221's caption preference — all additive, no disagreement between them.

frinsen:
Thanks @OpenClawLinda — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored — plus the 403 message now also names missing curl_cffi / JS runtime, and a refused YouTube media fetch retries through `mweb`/`tv`/`web_embedded` before degrading.

PR #231, opened 2026-09-19
Isolate setup/download tests from host binaries

charles98601-sg:
## Problem

Five tests fail on a clean checkout that doesn't happen to have `yt-dlp` installed — for reasons unrelated to what they assert:

```
FAILED tests/test_download.py::test_fetch_captions_requests_english_only
FAILED tests/test_download.py::test_download_url_requests_english_only
FAILED tests/test_setup.py::test_keyless_completed_setup_proceeds_silently
FAILED tests/test_setup.py::test_keyless_first_run_is_encouraged
FAILED tests/test_setup.py::test_key_present_is_ready
5 failed, 66 passed
```

Neither group actually needs a binary:

- **`test_download`** stubs `subprocess.run` to inspect the yt-dlp argv — but `download.py` bails at `shutil.which("yt-dlp")` *before* it builds that argv, so the stub is never reached.
- **`test_setup`** exercises the API-key state machine — but a missing binary short-circuits `--check` with exit 2 before the key logic runs. Whether the developer has ffmpeg installed is an unrelated variable for those cases.

## Change

Test-only; no production code touched.

- `conftest.py` gains a `fake_bin_dir` fixture providing no-op stand-ins on `PATH`. `test_setup` uses it via an autouse fixture, since `setup.py` runs as a subprocess and can't be monkeypatched directly.
- `test_download` stubs `shutil.which` alongside the existing `subprocess.run` stub.

## Verification

On a `PATH` containing ffmpeg but no yt-dlp:

| | result |
|---|---|
| before | 5 failed, 66 passed |
| after | **71 passed** |


frinsen:
Thanks @charles98601-sg — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored (the `fake_bin_dir` fixture now also writes `.bat` stubs on Windows for CI).

PR #232, opened 2026-09-19
Read .env through a single parser

charles98601-sg:
## Problem

There are three copies of the `.env` reader:

| location | handles inline comments |
|---|---|
| `config.read_env_file` | yes |
| `whisper.load_api_key._from_dotenv` | no |
| `setup._read_env_key` | no |

`83da59f` fixed the inline-comment handling in `config.py`, but the other two predate it, so the same line now resolves differently depending on who reads it.

Given a perfectly reasonable `~/.config/watch/.env`:

```
GROQ_API_KEY=sk-secret-abc   # my groq key
SETUP_COMPLETE=true   # finished on 2026-09-01
```

three symptoms fall out of the one root cause:

```
setup.py --json  ->  status         = ready
                     has_api_key    = True
                     setup_complete = False   # <- never registers
                     first_run      = True    # <- re-prompts forever

whisper.load_api_key() -> 'sk-secret-abc   # my groq key'
```

So `setup.py` reports everything is fine, the comment is sent as part of the bearer token and every transcription 401s, and because `SETUP_COMPLETE` never takes, the user is re-asked the first-run preference questions on every single run.

`setup.py` already imports `config` — it just also carried its own parser.

## Change

Promote the correct implementation to `config.read_env_value` (environment first, then each `.env` in order) and delete the other two. An `on_file` hook keeps setup's permission warning local to setup, where it belongs.

Net **-3 lines** of production code.

## Verification

Same `.env`, after:

```
setup_complete = True
first_run      = False
load_api_key() -> 'sk-secret-abc'
```

Quoted values keep a literal `#`, so `GROQ_API_KEY="sk-has#hash-inside"` survives unchanged.

12 regression tests added. `test_inline_comment_no_longer_reaches_the_api` fails against the previous `whisper.py` and passes after; the other new cases pass both ways, so they aren't over-asserting.

## Merge note

Touches `tests/test_whisper.py`. If `perf/parallel-whisper-chunks` lands first there's a trivial append conflict at the end of that file — both sides are independent test classes, keep both. Happy to rebase whichever order you prefer.


frinsen:
Thanks @charles98601-sg — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored — `read_env_value` is now the one parser for `config`, `setup`, `whisper` *and* the new `WATCH_COOKIES_*` / `WATCH_YTDLP` settings. #218's quoted-value-with-comment fix was folded in on top.

PR #233, opened 2026-09-19
Upload Whisper chunks concurrently

charles98601-sg:
## Problem

Audio over the 25 MB cap is split into chunks and uploaded one at a time, so total wait grows linearly with video length.

At 64 kbps mono each chunk holds roughly 50 minutes of audio, so a long podcast becomes several 24 MB POSTs that each block for tens of seconds. The uploads are independent — they were just queued behind each other.

## Change

Run them through a `ThreadPoolExecutor`. Wall clock is now bounded by the slowest chunk rather than their sum.

**Nothing user-visible changes.** Results are stitched in chunk order rather than completion order, so both the transcript and the per-chunk progress lines are byte-identical to a serial run. That's asserted directly:

```python
def test_progress_lines_match_a_serial_run(self, capsys):
    ...
    assert capsys.readouterr().err == (
        "[watch] chunk 1/3 → 1 segments\n"
        "[watch] chunk 2/3 failed — skipping (chunk 1 failed)\n"
        "[watch] chunk 3/3 → 1 segments\n"
    )
```

Partial-failure tolerance is unchanged: a chunk that exhausts its retries is still skipped, and only an all-chunk failure raises.

The cap is deliberately low (`MAX_PARALLEL_CHUNKS = 3`). These APIs rate-limit and `_post_whisper` gives up after `MAX_429_RETRIES`, so saturating the endpoint would trade wall clock for lost chunks.

## Verification

Stubbed uploader, 4 chunks, same code path at different worker counts:

| workers | wall clock | |
|---|---|---|
| 1 | 1.23s | (equivalent to previous behaviour) |
| 3 | 0.61s | default |
| 4 | 0.31s | **3.9x** |

stderr byte-identical across all three. Against a real ~45s/chunk upload that's roughly 180s → 45s.

Four new tests cover overlap, the worker cap, out-of-order completion still stitching correctly, and the output-parity assertion above. The three existing `TestTranscribeChunks` cases pass unchanged.

## Merge note

Touches `tests/test_whisper.py`; see the same note on `fix/single-env-parser` — trivial append conflict if that one lands first, keep both classes.


frinsen:
Thanks @charles98601-sg — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored, with your byte-identical-progress test intact.

PR #234, opened 2026-09-19
Fall back to the video's own caption language

charles98601-sg:
## Problem

Captions are requested as `en.*` only, so a video without English subtitles comes back with nothing and falls through to a Whisper upload — slower, and billed — even when the platform has perfectly good captions in the language actually spoken.

`SKILL.md` already describes this step as *"pulls manual or auto-generated subtitles from the source platform if available"*, so the implementation was narrower than the documented contract.

## Change

English is still requested first, so **English videos are unchanged at one round trip**. Only when that returns nothing does the script read the `info.json` it just wrote — `subtitles` and `automatic_captions` are part of yt-dlp's documented info-dict contract — and retry against exactly one tag:

- an uploader-supplied track if one exists (skipping `live_chat`), else
- the `-orig` automatic track, which is the spoken language rather than a machine translation of it.

**The bound this module was built around still holds.** The retry requests a single tag and never widens to `all`, which on YouTube means hundreds of translated tracks and a multi-minute stall — exactly what the module docstring warns about. That's asserted directly:

```python
def test_retry_never_requests_all(monkeypatch, tmp_path):
    # info.json advertises 300 automatic_captions tracks
    for call in calls:
        assert "all" not in _sub_langs(call).split(",")
```

`download_url` reads the same `info.json`, which `fetch_captions` has already written into that directory, so the discovered language costs no extra extraction.

## Verification

8 new tests: English stays single-round-trip, non-English retries exactly once with the right tag, manual beats auto, `-orig` beats translations, `live_chat` is not a caption track, no tracks means no retry, and the never-`all` guard above.

Five of them fail against the current `download.py`; the three asserting unchanged behaviour pass both ways.

**Not verified end-to-end against a live non-English video** — YouTube now requires cookies for anonymous extraction from this host. The behaviour is covered by unit tests built on the documented info-dict contract, matching how the existing yt-dlp argv tests are written. Happy to adjust if you know of a platform worth pinning a live test against.

## Merge note

Touches `download_url`, as does `feat/cache-downloads`. The two changes are orthogonal within that function (cache check at the top, language detection at argv construction) and I've verified they compose — tell me which lands first and I'll rebase the other.


frinsen:
Thanks @charles98601-sg — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): adapted — combined with #212 (video's `language` field, since YouTube serves a translated `en` track rather than nothing) and #221 (human > auto ranking) into one implementation; your `-orig` fallback for an unknown language and the "never `all`" guard are in it verbatim.

PR #235, opened 2026-09-19
Cache downloads so a repeat run skips the fetch

charles98601-sg:
## Problem

Downloading dominates a run. The README's own measurement is **~37s / 76 MB** for a clip whose frames then take about a second to extract — and every invocation pays it again, including a follow-up question about a video just watched.

`SKILL.md` already works around this by hand, and the workaround contradicts itself:

> **Step 5 — clean up.** [...] delete it with `rm -rf <dir>`

> 3. Re-run with `--timestamps ...`. For a URL, point the second run at the **downloaded local file** in the work dir so it doesn't re-download.

One tells the model to delete the directory, the other tells it to reach back into that same directory. Neither survives a new session.

## Change

Key the download by URL under `~/.cache/watch/downloads` (honouring `XDG_CACHE_HOME`) and reuse it.

Audio-only and full-video fetches get **separate entries**, so a `transcript` run can never hand a later frame run a file with no video stream.

Bounded by construction:

- 2 GB cap, least-recently-used evicted after each download
- the entry the current run needs is exempt from eviction
- a zero-byte file from an interrupted run counts as a miss rather than being served
- `--no-cache` opts out, and so does `--out-dir` — naming a directory means the user wants everything in that one place

`SKILL.md` updated: the contradictory manual workaround is gone, step 5 now notes that the work dir holds frames while the download lives in the cache, and the security section states what is written where.

## Verification

End-to-end against a local HTTP server (no external platform involved):

| run | wall clock | |
|---|---|---|
| 1st | 2.49s | downloaded |
| 2nd | 0.90s | **cache hit** |
| 3rd `--no-cache` | 3.84s | downloaded, correctly bypassing |

Both runs exit 0 and the **generated reports are byte-identical** (modulo the randomly-named work dir). A `--detail transcript` run leaves the video entry untouched, and a subsequent frame run correctly picks the video entry rather than the audio one. Over a real network the saving is the full download time.

14 new unit tests cover key separation and stability, hit/miss behaviour, the zero-byte guard, `use_cache=False`, and LRU eviction including the `keep` exemption.

## Merge note

Touches `download_url`, as does `feat/native-language-captions`; see the note on that PR. I've merged all five branches locally to confirm they compose (106 tests pass on the combined tree) — happy to rebase in whatever order suits you.


robinjose911:
Thanks for this — well-argued, and the point about SKILL.md contradicting itself is correct. Measured on a 16-minute video: 16.2s cold, 8.7s warm, identical frame selection both runs.

Two things you couldn't have known about, both of which bit when this is combined with #221 (human-authored caption preference), in case they're useful if either lands:

**1.** `_cached_download` calls `_pick_subtitle(out_dir)`. Once #221 adds the manual-language argument, every cache **hit** silently drops the human-authored preference and falls back to auto ranking — a manual track sitting in the same directory gets ignored. Passing `_manual_sub_langs(info_path)` there fixes it.

**2.** The cache key is URL + audio/video. It doesn't include the media format, so changing the resolution cap — e.g. lifting the hardcoded `height<=720`, which #215 proposes — serves the cached 720p file for that URL indefinitely, with a completely healthy-looking run and no way to notice. Folding the format spec into the key makes it miss cleanly and lets LRU reclaim the old entry.

Neither is a flaw in this PR on its own; they're interaction bugs that only appear once other work lands. Both were live after cherry-picking both PRs cleanly.

Ported into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `a859310`, fixes in `ebfaa5e`) with your authorship preserved, credited in CREDITS.md. Would still rather see this merged here.

PR #121, opened 2026-08-07
feat(watch): data-mining track — transcript verification, vision extraction, scene docs

noblecloud:
## What

Adds a data-mining track to `/watch` — turning a watched video into verified, structured, timecoded knowledge instead of just answering a question.

**Audible track — `verify.py`**: runs Whisper STT against the downloaded transcript, aligns both SRTs (text-first within an 8s window, since whisper timestamps drift), flags low-similarity segments as `candidates.json` for an LLM verdict pass, and emits a provisional `aligned.srt`. Fixes the caption-first blind spot: auto-captions are frequently wrong on exactly the terms that matter (numbers, names, technical terms).

**Visual track — `vision_extract.py`**: sends extracted keyframes to a vision model and writes `index.json` with per-frame `{timecode, type, text, graphics}` — on-screen text verbatim, graphical content described, broll/face auto-skipped. Threaded, retry-safe.

**Combination — `scene_doc.py`**: merges the verified transcript + frame index into a timecoded scene document with scene directions and inline graphical content — ready for knowledge-base import.

## Notes

- Pure stdlib (difflib, urllib, threading) — no new dependencies.
- Model routing: cheap text model for all LLM work; vision model only for frame analysis.
- Tested against a live run: `vision_extract.py` required a browser User-Agent to avoid 403s on the OpenAI-compatible endpoint and `temperature=1` for kimi-k2.7-code (both included in this PR).

Issue #66, opened 2026-07-11
[Wish] Can I use this tool to fix the machine-transcription (mishear)?


enihcam:
I've always been looking for a tool to help me fix the mishear issue in machine-transcription - to make the captions more scene-aware or context-aware. So I wish there's a tool like claude-video that could pull the timestamped captions (manual or auto-generated) only as the target to be fixed/refined. Is it feasible?

amarakramali:
Yes — the current pipeline already has most of the extraction and visual-grounding pieces, but it does **not** yet write a corrected VTT/SRT file.

A practical two-pass workflow today would be:

1. Run `/watch <url> --detail transcript` and explicitly ask for the raw timestamped transcript plus a list of suspicious segments (proper nouns, homophones, abrupt context changes, etc.). In this mode, captioned URLs skip the video download and use the VTT pulled by `yt-dlp`.
2. For the suspicious timestamps, re-run against the downloaded/local video with something like:

   ```text
   /watch <local-video> --detail transcript --timestamps 1:23,4:56,7:08 --no-whisper
   ```

   With `transcript` detail, those become the **only** frames extracted. Claude can then compare each questionable line with the relevant visual plus the neighboring transcript context, without spending tokens on scene sampling across the whole video.
3. Apply only the proposed text replacements to the original `video*.vtt` in the reported work directory, preserving its cue timing.

That last step is the important current limitation. `transcribe.py` parses VTT cues into `{start, end, text}` segments, deduplicates rolling YouTube captions, and the displayed transcript only prints `[MM:SS]`. Therefore a future write-back feature should edit the original cue blocks (or retain cue IDs/raw timing explicitly) rather than regenerate VTT from the formatted transcript, otherwise it could lose millisecond timing and rolling-cue structure.

A safe built-in feature could look like `--refine-captions output.vtt` with these constraints:

- never overwrite the source caption file;
- preserve cue start/end times by default and change text only;
- use a small neighboring-cue window plus a frame at each suspect cue midpoint;
- emit a reviewable before/after diff and confidence/reason for every change;
- leave uncertain audio-only corrections untouched unless the visual/context evidence is strong.

So: **feasible now as a guided two-pass workflow, and feasible as a focused feature**, but automatic caption-file export is not implemented yet.

PR #236, opened 2026-09-19
docs(watch): flag transcript proper nouns as unverified

gth-spec:
## Problem

Captions and Whisper both transcribe phonetically, so proper nouns come back wrong or invented. Watching a 28-minute interview this week, the auto-captions returned "Dooo Almeida" for Diogo Almeida, "Verscell" for Vercel, and "chatbt" for ChatGPT — in a transcript that was otherwise accurate and carried no hallucination signal at all.

Nothing in SKILL.md tells the model to doubt those. The transcript is presented as one of two streams of evidence, so a misheard name is restated to the user as fact, and from there it gets copied into notes, quoted, or searched for. The failure is quiet: unlike a garbled sentence, a garbled name reads as perfectly fluent.

This is the "caption-first blind spot" named in the rationale for #121 — *"auto-captions are frequently wrong on exactly the terms that matter (numbers, names, technical terms)"* — and the mishearing problem raised in #66.

## Why this and not the existing work

- **#121** (closed, unmerged) proposed solving it with a data-mining track: Whisper-vs-caption alignment, vision extraction, scene docs, four new scripts. This PR is the lightweight counterpart — no new scripts, no new dependencies, no runtime cost.
- **#66**'s answer is a guided three-pass workflow the user drives *after* they already suspect a name is wrong. This makes the model flag it unprompted, which is the case that actually bites.
- **#223** (open) flags whole transcripts that look hallucinated via `no_speech_prob`. Complementary, not overlapping: that catches Whisper inventing dialogue over music or silence; this catches individual names inside a transcript that is otherwise correct, including native captions, where no confidence signal exists.

## Change

One paragraph in Step 4 of `skills/watch/SKILL.md`. It says to check a name against a frame before stating it — title cards, slides and on-screen UI usually spell it correctly — and where no frame confirms it, to give the name as-heard and mark it unverified.

It covers the `transcript` detail case explicitly, since that mode has no frames at all and the guidance would otherwise be unactionable there.

Scope is named deliberately: people, companies, products, model names, URLs, prices and figures. Anything the user might repeat, quote or act on.

## Notes

- Docs-only. No script changes, no test changes, no version bump.
- Leaves the existing Step 4 guidance untouched, including the `transcript`-detail summary paragraph.
- Does not reintroduce `${CLAUDE_SKILL_DIR}`, move `SKILL.md`, or add a `commands/` wrapper (per AGENTS.md).
- Every example is a real garble observed in one live run, not invented for illustration.

---


robinjose911:
Thanks for this — it's two lines and it's the highest-value change I've read in this repo.

A data point that backs it harder than the PR claims. I checked a real auto-caption track against a summary that had been written from that video. The words "TypeSafe", "Diogo", "Almeida" and "Vercel" appear **zero times** in the captions — not misspelled, simply absent — yet all four were central enough to end up in the summary. They were legible only from frames: a tweet screenshot and a chart legend. A `--detail transcript` run could not have produced those names at all, correctly or otherwise.

So the failure mode is worse than phonetic mangling. A name can be confidently stated while having no basis in the transcript whatsoever, and nothing in the report signals that.

Cherry-picked into a personal fork ([robinjose911/claude-video](https://github.com/robinjose911/claude-video), commit `6171c99`) with your authorship preserved, credited in CREDITS.md, since this repo has been inactive since June. Would still rather see it merged here.

frinsen:
Thanks @gth-spec — this is in the community fork [frinsen/claude-video](https://github.com/frinsen/claude-video) as of [v0.3.0](https://github.com/frinsen/claude-video/releases/tag/v0.3.0): merged as authored.

